PR #348, opened 2014-07-23
bugfix of raw_command

dheule:
with raw_command, you can set env variables in crontab,
but cron don't accept env vars if the definition is
like this:
SHELL=/bin/bash # marker of something ...
so we need to make sure that raw commands are not apended by a marker

### DIFF
--- a/cdist/conf/type/__cron/gencode-remote
+++ b/cdist/conf/type/__cron/gencode-remote
@@ -26,7 +26,7 @@ command="$(cat "$__object/parameter/command")"
 
 if [ -f "$__object/parameter/raw" ]; then
    raw="$(cat "$__object/parameter/raw")"
-   entry="$raw $command"
+   entry="$raw $command # $name"
 elif [ -f "$__object/parameter/raw_command" ]; then
    entry="$command"
 else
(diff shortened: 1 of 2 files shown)


dheule:
Any news on this ?
For example vixie-cron is refusing the whole crontab if  a #someting marker is there ...

Please review and pull

cdist/conf/type/__cron/gencode-remote
```diff
@@ -26,7 +26,7 @@ command="$(cat "$__object/parameter/command")"
 
 if [ -f "$__object/parameter/raw" ]; then
    raw="$(cat "$__object/parameter/raw")"
-   entry="$raw $command"
+   entry="$raw $command # $name"
```
darko-poljak: @telmich this '# $name' is here because it is appended for every case except for raw_command. So it should be correct.

darko-poljak:
@telmich @dheule I reviewed the changes and I think the changes are correct.
For every case type appends marker comment, except for raw_command case.
@dheule I assume you have tested all this.
I always remember Knuths's:
_Beware of bugs in the above code; I have only proved it correct, not tried it._
I will merge this. If it proves wrong we can revert it :)

PR #352, opened 2014-08-12
Expand the documentation of TYPE execution inside the shell

tpo:
I've added an explanation of how to exactly execute TYPEs in the shell. Please pull

Greets!
*t


darko-poljak:
Broken into 354 and 355.

PR #354, opened 2014-08-18
Please merge __package_dpkg type

tpo:
Dear Nico,

as allready discussed via PM, please add the __package_dpkg type which lets you install local Debian packages from files/\* on the target.

Comments wellcome, thanks,
*t


darko-poljak:
@telmich Can you review this one?
@tpo Can you migrate man.text to reStructuredText man.rst ?

tpo:
@darko-poljak as requested, I've moved to man.rst, as you should be able to see in commit 8e3281a. Nothing else has been touched in that commit.

tpo:
@darko-poljak : With commit 6bcfdec I've also removed the reference to cdist_type and moved to GPL3+ as you've requested here: https://github.com/ungleich/cdist/pull/360#issuecomment-264110087

darko-poljak:
@tpo Just one more '=' in man.rst header. Sorry for torturing you.

darko-poljak:
@tpo How about changing this type so that you don't have to specify both full path and deb file basename?
For example, __object_id is path to deb file
__package_dpkg $__type/files/bar_1.4.deb
and then you can use `basename` to get file's name.
And what do you think about allowing future type extensions by adding an `action` parameter which if not specified defaults to install? As is now, this type supports only install action.

tpo:
Hi @darko-poljak 

I wrote __package_dpkg and __postgres_extension while working on a project that was using cdist. Since I am not using cdist these times I will not invest a lot of time into this. Patching the documentation was easy so I did that for you, but changing the interface requires more work.

Changing the interface as you propose should be upwards compatible and can be done later if anybody is willing to do the effort.

tpo:
I've added the '=' to the title as you've requested in commit 4370efd

darko-poljak:
@tpo Aha, I see, ok. Thanks for this additional work.
@telmich Fine with you to merge this PR?

PR #355, opened 2014-08-18
Please pull: clarify cdist_type cli execution

tpo:
Hello Nico,

here's a little docu improvement. Please pull. Comments wellcome. Thanks,
*t


darko-poljak:
@tpo This change removes line for type usage in init and type manifest and describes only usage in cdist shell. I will close this PR and you can create another one with both, manifest and cdist shell usage, if you think it needs more clarification.

PR #356, opened 2014-08-19
add support for 'replication' right for PostgreSQL roles

tremlin:
Since version 9.1 PostgreSQL supports the REPLICATION right: http://www.postgresql.org/docs/9.1/static/sql-createrole.html


darko-poljak:
@tremlin Can you migrate man.text to reStructuredText man.rst format? Then we will review this PR again.

darko-poljak:
Nobody replies, no update since 2014 and there are conflicting files.
If someone needs this, the change is really simple so it should be very easy to create new pull request.

PR #366, opened 2014-10-02
split cache into separate dirs per start time

dheule:
I have chosen this structure to simplify the cleanup
of the old configure runs. Now the structure is:
cdist/cache/yyyy-mm-dd/HHMMSS/hostname
the time is the starttime of the configure run,
so if you configure your host serial, you have more than one
directory at the time level, if you start with -p, you have one.

### DIFF
--- a/cdist/config.py
+++ b/cdist/config.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 #
 # 2010-2013 Nico Schottelius (nico-cdist at schottelius.org)
+# 2014      Daniel Heule     (hda at sfs.biz)
 #
 # This file is part of cdist.
 #
@@ -155,7 +156,7 @@ class Config(object):
         self.manifest.run_initial_manifest(self.local.initial_manifest)
         self.iterate_until_finished()
 
-        self.local.save_cache()
+        self.local.save_cache(start_time)
         self.log.info("Finished successful run in %s seconds", time.time() - start_time)
 
 
--- a/cdist/exec/local.py
+++ b/cdist/exec/local.py
@@ -2,6 +2,7 @@
 #
 # 2011 Steven Armstrong (steven-cdist at armstrong.cc)
 # 2011-2013 Nico Schottelius (nico-cdist at schottelius.org)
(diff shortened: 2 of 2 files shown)


4nd3r:
what happened with this PR?

I would like to do some (recursive) grepping and/or log checking of last run afterwards, so cache structure `yyyy-mm-dd/HHMMSS/hostname` would be really nice addition.

darko-poljak:
@telmich @asteven @dheule There was no agreement on this. Can we discuss about this all over again and have one? :) Whether it be that we do cache dir split and how or that we don't.

4nd3r:
can we make cache structure configurable?

darko-poljak:
You mean we add an option for specifying cache path format?

If not specified then default, `<cache-dir>/<hostdir>` is used.
If specified then specified format is used. We can support subdirectories and placeholders,
%h for hostdir (we can require this in format), %Y, %m, %d, %H, %M, %S (datetime), %p (pid), %u (uid), etc.
Then if format is `"/%h-%Y-%m-%d_%H_%M_%S-%p"` cache dir is:
`.cdist/cache/<hostdir>-2016-01-01_18_34_01-34567`
(without leading '/' it would be the same).
For `"%Y-%m-%d-%H-%M-%S-%p/%h"` it will be:
`./cdist/cache/2016-01-01-18-34-01-34567/<hostdir>`.

dheule:
The ability to configure via parameters how the cache dir is structured
sounds relay good to me.
So everyone can configure for his needs.

I think this could also be done via a env var which can be set in the
inital manifest, so no need to specify a parameter for each run to be
consisent ...

Thank you for working on this ...

Daniel


2016-11-29 12:30 GMT+01:00 Darko Poljak <notifications@github.com>:

darko-poljak:
@dheule Do you agree that we close this PR? And I will continue working on cache format option/env var and create new PR when I am done.

dheule:
For me it's fine to close this old PR.

Thank you,
Daniel

2016-11-29 14:38 GMT+01:00 Darko Poljak <notifications@github.com>:

PR #368, opened 2014-10-05
Set from __locale type, the system-wide default

jimenezrick:


telmich:
Hey @jimenezrick !

We have merged __locale_system recently - you can use that for setting the system default!

PR #382, opened 2014-12-31
Cdist fails silently when there's a directory in __type/parameter/default

acatton:
I just spent one hour on this bug. I couldn't figure out why cdist was ignoring my default parameters.

After investigating on the bug myself, and thanks to the code being very well structured, I found out where was the bug. I wrote a test case, and fixed it.

I assumed that cdist shouldn't be bothered if there's a subdirectory in `__type/parameter/default/`, and it should just keep carrying its own business. Please let me know if this behavior is not the one you want, i'll change my fix.

Thank you.


cdist/core/cdist_type.py
```diff
@@ -204,11 +204,13 @@ def parameter_defaults(self):
...
                 defaults_dir = os.path.join(self.absolute_path, "parameter", "default")
                 for name in os.listdir(defaults_dir):
-                    with open(os.path.join(defaults_dir, name)) as fd:
-                        defaults[name] = fd.read().strip()
+                    try:
+                        with open(os.path.join(defaults_dir, name)) as fd:
+                            defaults[name] = fd.read().strip()
+                    except EnvironmentError:
+                        pass  # Swallow errors raised by open() or read()
             except EnvironmentError:
```
telmich: Hello @acatton! First of all thanks for your PR. 

I am returining to work from a short break and try to wrap my head around what you did. Afaics, the outer try...except block catches the same exception as you have done in the inner, so your change should not make a difference.

Can you explain why this prevents the problem from happening?

I wonder if it makes sense to actually stat() before and only read the file, if it is a normal file and then to read it.

Btw, I have just seen that EnvironmentError vanished in python 3.3:

Changed in version 3.3: EnvironmentError, IOError, WindowsError, VMSError, socket.error, select.error and mmap.error have been merged into OSError.
from https://docs.python.org/3/library/exceptions.html

acatton: Hello @telmich,

Sure, I do understand it could be confusing at first. I hope this will clarify it (let me know if it's still blurry :P ):

Originally the `try ... except EnvironmentError` catches the following errors:
 * `os.listdir(defaults_dir)` where `defaults_dir` doesn't exist
 * `open(file)` where `file` is actually a directory
 * `fd.read()` where `fd` is not readable by the user
 * (other errors that I'm not aware of on `listdir()`, `open()` and `read()`)

The bug I encountered was when there's a directory will be given to `open()`, because the original `try ... except EnvironmentError` is outside the loop, it stops iterating, and keep going.
I added an inner `try ... except` block so that if an `open()` or a `read()` fails for any reason, the loop would keep going. (I added tests for this) And I kept the outer `try ... except` for `os.listdir()`.

Does it make sense?

For the `EnvironmentError`, I didn't notice that, since [`EnvironmentError` was kept for backward compatibility](https://docs.python.org/3/library/exceptions.html#EnvironmentError). If you want to be compatible python 3.2 we should keep `EnvironmentError`, otherwise I think we should rename it to `OSError`.

Let me know if you want me to modify/fix my pull request in any way.

Thank you.

acatton:
@telmich I just realized my test was wrong. Git (as I forgot) don't track empty directories. I rebased my pull request, you should be able to reproduce the error locally if you `git checkout 9717d5d`.

telmich:
Thanks, merged!

Issue #395, opened 2015-05-08
__user_groups doesn't work on SELS11


dheule:
gpasswd -a isn't known on older linuxes like SLES11, so __user_groups doesn't work on this ones.
I submit a bugfis pull request shortly

PR #396, opened 2015-05-08
fix for older linuxes where gpasswd doesn't have the relevant options…

dheule:
fixes Issue #395

### DIFF
--- a/cdist/conf/type/__user_groups/gencode-remote
+++ b/cdist/conf/type/__user_groups/gencode-remote
@@ -21,6 +21,20 @@
 user="$(cat "$__object/parameter/user" 2>/dev/null || echo "$__object_id")"
 state_should="$(cat "$__object/parameter/state")"
 
+os_version="$(cat "$__global/explorer/os_version")"
+case "$os_version" in
+    SUSE\ Linux\ Enterprise\ Server\ 11*VERSION\ =\ 11*)
+        old=yes
+        addparam="-A"
+        delparam="-R"
+    ;;
+    *)
+        old=no
+        addparam="-a"
+        delparam="-d"
+    ;;
+esac
+
 mkdir "$__object/files"
 # file has to be sorted for comparison with `comm`
 sort "$__object/parameter/group" > "$__object/files/group.sorted"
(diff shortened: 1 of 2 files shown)


telmich:
I wonder - should we user usermod in any case instead of gpasswd? @asteven what do you think?

I think it is an appropriate patch, but I try to minimise the number of cases for special OS, if possible.

dheule:
usermod has also different incompatible parameter Syntax from old to new
linuxes ;-((
Am 08.05.2015 15:54 schrieb "Nico Schottelius" <notifications@github.com>:

dheule:
I had an idea, pleas wait for the next update off the pullrequest. Tank
you. Daniel
Am 08.05.2015 15:54 schrieb "Nico Schottelius" <notifications@github.com>:

dheule:
Now i have implemented an explorer for old usermod implementations.
the switch from usermod and gpasswd is relevant because new usermod commands 
have the ability to append some groups to a user, but i don't have found an option 
to remove only one group from a user, so in the old world we must use usermod, in the newer ones we must use gpasswd.
But now we have no more a hardcoded os'es ...

Please review ...

cdist/conf/type/__user_groups/gencode-remote
```diff
@@ -20,6 +20,17 @@
...
 user="$(cat "$__object/parameter/user" 2>/dev/null || echo "$__object_id")"
 state_should="$(cat "$__object/parameter/state")"
+oldusermod="$(cat "$__object/explorer/oldusermod")"
+
+if [ "$oldusermod" = "true" ]; then
+        old=yes
+        addparam="-A"
+        delparam="-R"
+else
+        old=no
```
telmich: Minor improvement possible: if we have $oldusermod, we do not need to setup old=, but can use $oldusermod below again :-)

dheule: Done ...

2015-05-11 10:14 GMT+02:00 Nico Schottelius <notifications@github.com>:

dheule: Ping ?

2015-05-11 10:19 GMT+02:00 Daniel Heule <daniel.heule@gmail.com>:

dheule:
hello nico,
is something against a merge of this fix in the current version ?

please comment ;-)

telmich:
Everything fine :-)

PR #405, opened 2015-10-15
Treat NetBSD the same as FreeBSD and OpenBSD in 'stat' explorers

jakllsch:


telmich:
Great size of patches - that's really good for merging!

PR #406, opened 2015-10-15
use passwd database instead of shadow database on NetBSD too

jakllsch:


telmich:
Sorry for the long delay, we have been very busy!

PR #411, opened 2016-02-23
Added Mitel as a redhat derived OS

sleemburg:
Hi,

We have some Mitel PABX machines which are a Red Hat derivative, but use different version numbering. This extra os explorer may be helpful for others too.


telmich:
Thanks a lot, Stephan!

PR #414, opened 2016-03-06
Add ssh multiplexing options to default REMOTE_EXEC and REMOTE_COPY.

darko-poljak:
Hi!

I added ssh multiplexing options to default REMOTE_EXEC and REMOTE_COPY.

### DIFF
--- a/scripts/cdist
+++ b/scripts/cdist
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 #
 # 2010-2013 Nico Schottelius (nico-cdist at schottelius.org)
+# 2016 Darko Poljak (darko.poljak at gmail.com)
 #
 # This file is part of cdist.
 #
@@ -19,6 +20,32 @@
 # along with cdist. If not, see <http://www.gnu.org/licenses/>.
 #
 #
+
+def inspect_ssh_mux_opts():
+    """Inspect whether or not ssh supports multiplexing options"""
+    wanted_mux_opts = {
+        "ControlPath":"~/.ssh/master-%l-%r@%h:%p",
+        "ControlMaster":"auto",
+        "ControlPersist":"125",
+    }
+    # if checked key option is present then this assumes
+    # all options in value are present
+    check = {
+        "ControlMaster": ("ControlMaster", "ControlPath"),
+        "ControlPersist": ("ControlPersist",),
+    }
+    mux_opts = {}
+    for x in check:
+        try:
+            subprocess.check_output("ssh -o {}".format(x),
+                    stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            foo = e.output.decode().lower()
+            if not "bad configuration option" in foo:
+                for o in check[x]:
+                    mux_opts[o] = wanted_mux_opts[o]
+    foo = ["-o {}={}".format(x, mux_opts[x]) for x in mux_opts]
+    return " ".join(foo)
 
 def commandline():
     """Parse command line"""
@@ -53,6 +80,11 @@ def commandline():
     parser['banner'].set_defaults(func=cdist.banner.banner)
 
     # Config
+    # inspect multiplexing options for default remote copy/exec scp/ssh
+    MUX_OPTS = inspect_ssh_mux_opts()
+    cdist.REMOTE_EXEC += MUX_OPTS
+    cdist.REMOTE_COPY += MUX_OPTS
+
     parser['config'] = parser['sub'].add_parser('config',
         parents=[parser['loglevel']])
     parser['config'].add_argument('host', nargs='+',
--- a/scripts/cdist
+++ b/scripts/cdist
@@ -22,6 +22,8 @@
 #
 
 def inspect_ssh_mux_opts():
+    import subprocess
+
     """Inspect whether or not ssh supports multiplexing options"""
     wanted_mux_opts = {
         "ControlPath":"~/.ssh/master-%l-%r@%h:%p",
@@ -38,13 +40,13 @@ def inspect_ssh_mux_opts():
     for x in check:
         try:
             subprocess.check_output("ssh -o {}".format(x),
-                    stderr=subprocess.STDOUT)
+                    stderr=subprocess.STDOUT, shell=True)
         except subprocess.CalledProcessError as e:
             foo = e.output.decode().lower()
             if not "bad configuration option" in foo:
                 for o in check[x]:
                     mux_opts[o] = wanted_mux_opts[o]
-    foo = ["-o {}={}".format(x, mux_opts[x]) for x in mux_opts]
+    foo = [" -o {}={}".format(x, mux_opts[x]) for x in mux_opts]
     return " ".join(foo)
 
 def commandline():
--- a/cdist/__init__.py
+++ b/cdist/__init__.py
@@ -41,8 +41,10 @@ BANNER = """
    "P'        ""         ""
 """
 
-REMOTE_COPY = "scp -o User=root -q"
-REMOTE_EXEC = "ssh -o User=root -q"
+SSH_MUX_OPTS = ("-o ControlPath=~/.ssh/master-%l-%r@%h:%p"
+    " -o ControlMaster=auto -o ControlPersist=125")
+REMOTE_COPY = "scp -o User=root -q " + SSH_MUX_OPTS
+REMOTE_EXEC = "ssh -o User=root -q " + SSH_MUX_OPTS
 
 class Error(Exception):
     """Base exception class for this project"""
--- a/cdist/__init__.py
+++ b/cdist/__init__.py
@@ -41,10 +41,8 @@ BANNER = """
    "P'        ""         ""
 """
 
-SSH_MUX_OPTS = ("-o ControlPath=~/.ssh/master-%l-%r@%h:%p"
-    " -o ControlMaster=auto -o ControlPersist=125")
-REMOTE_COPY = "scp -o User=root -q " + SSH_MUX_OPTS
-REMOTE_EXEC = "ssh -o User=root -q " + SSH_MUX_OPTS
+REMOTE_COPY = "scp -o User=root -q"
+REMOTE_EXEC = "ssh -o User=root -q"
 
 class Error(Exception):
     """Base exception class for this project"""
--- a/cdist/exec/local.py
+++ b/cdist/exec/local.py
@@ -195,7 +195,7 @@ class Local(object):
         except subprocess.CalledProcessError:
             raise cdist.Error("Command failed: " + " ".join(command))
         except OSError as error:
-            raise cdist.Error(" ".join(*args) + ": " + error.args[1])
+            raise cdist.Error(" ".join((str(x) for x in error.args)))
         finally:
             if message_prefix:
                 message.merge_messages()


asteven:
Hi

Thanks for the contribution.
Your patch makes assumption about where the control master socket is stored. This will not work for everybody. Also ControlPersist is not supported in older ssh versions, e.g. on RHEL 6.

IMHO the better and more flexible way to solve this is to configure it in ~/.ssh/config or create a wrapper script for ssh/scp.

Cheers

telmich:
@asteven I was suggesting to add this to cdist and save the socket in the temporary directory. You are right that this breaks on very old config hosts (rhel6 ;-), but I think it's better to include this in cdist rather than letting everybody redoing it.

Maybe we should settle on something that checks the ssh version and includes it by default, if available?

jdguffey:
@telmich: I'm inclined to agree with @asteven. In my own situation, I have a particular configuration already in place via ~/.ssh/config that I'd like to use whenever I'm trying to work on machines. Perhaps parsing the configuration or the wrapper script would be best?

@telmich, @asteven, @darko-poljak: Forgive my lack of understanding, but as I read through this, I just see that it *can* make use of connection multiplexing, not necessarily that it *does*. Am I correct in this? What I mean is, how does the initial connection stay open for later muxing? In the past whenever I've executed cdist, in order to achieve multiplexing, I've had to hold a connection open aside from cdist, execute cdist, then close the other connection afterward. If I didn't do this, cdist would connect/disconnect for every remote copy or execute operation. Is that what ControlPersist is for?

asteven:
Jake Guffey wrote on 03/07/16 18:06:
> @telmich: I'm inclined to agree with @asteven. In my own situation, I have a particular configuration already in place via ~/.ssh/config that I'd like to use whenever I'm trying to work on machines. Perhaps parsing the configuration or the wrapper script would be best?

I use a wrapper script like this

```sh
[8:38:39] crius:~% ls -al ~/.cdist-hpc/bin
total 22
drwxr-xr-x 2 sar sar   4 Mar  8 08:38 .
drwxr-xr-x 7 sar sar  13 Feb 23 23:20 ..
lrwxrwxrwx 1 sar sar   3 Sep  8  2013 scp -> ssh
-rwxr-xr-x 1 sar sar 920 Jul 22  2015 ssh

```

```sh
[8:40:05] crius:~% cat ~/.cdist-hpc/bin/ssh
#!/bin/sh
#
# Wrapper to set cdist specific ssh/scp options
#

my_name="${0##*/}"

#echo "$my_name: $@" | logger

: ${XDG_RUNTIME_DIR:="$(d="/tmp/xdg-runtime-$USER"; umask 077 && mkdir 
-p "$d" && echo "$d")"}

control_path_base_dir="$XDG_RUNTIME_DIR/cdist-ssh"
[ -d "$control_path_base_dir" ] || {
    echo "control_path_base_dir: $control_path_base_dir" | logger
    mkdir "$control_path_base_dir"
    chmod 700 "$control_path_base_dir"
}
target_host_id="$(echo "${__target_host##*+}" | cksum | awk '{print $1}')"
control_path_dir="$control_path_base_dir/${target_host_id}"
[ -d "$control_path_dir" ] || {
    echo "control_path_dir: $control_path_dir" | logger
    mkdir "$control_path_dir"
}

# Note: the bloddy stinkin redhat 6 does not support ControlPersist :-/
exec /usr/bin/$my_name \
    -o ForwardX11=no \
    -o ControlMaster=auto \
    -o ControlPath="$control_path_dir/%r@%h:%p" \
    -o ControlPersist=10 \
    $@
```

>
> @telmich, @asteven, @darko-poljak: Forgive my lack of understanding, but as I read through this, I just see that it *can* make use of connection multiplexing, not necessarily that it *does*.
> Am I correct in this? What I mean is, how does the initial connection stay open for later muxing?
> In the past whenever I've executed cdist, in order to achieve multiplexing, I've had to hold a connection open aside from cdist, execute cdist, then close the other connection afterward. If I didn't do this, cdist would connect/disconnect for every remote copy or execute operation. Is that what ControlPersist is for?

This is a feature of ssh 6.* (not 100% sure about the version. I know 
that 5.3 on RHEL 6 does not support it). In older ssh versions you would 
get the symptoms you describe.

asteven:
Nico Schottelius wrote on 03/07/16 17:13:
> @asteven I was suggesting to add this to cdist and save the socket in the temporary directory. You are right that this breaks on very old config hosts (rhel6 ;-), but I think it's better to include this in cdist rather than letting everybody redoing it.

Whatever you include will be wrong for at least some people. The 
proposed patch will break cdist for me, I'd have to revert it in my copy.

>
> Maybe we should settle on something that checks the ssh version and includes it by default, if available?

I do not see the point. IMHO there is no unsolved problem to solve here. 
It's easy to configure this or wrap ssh/scp to do whatever you like.

If you really feel there is a need for another way to deal with this 
then use an environment variable to override the default, e.g. 
CDIST_REMOTE_{EXEC,COPY}.

telmich:
The problem with any external resource (like wrapper scripts) is that fixes it only for you, not for everyone.

If we include automatic selection based on the ssh version and save the control master in the temporay cdist directory, this should not affect anyone with older ssh versions (as in that case the options will not be set), but everyone with a recent version benefits from it.

More than 80% of the users benefit from the proposed change and as far as I can see, there are no real users who have a disadvantage when we enable automatic enabling based on the ssh version.

darko-poljak:
I searched for a way to determine ssh version.
As for e.g. in [http://linux.101hacks.com/ssh-client-commands/ssh-client-version/](http://linux.101hacks.com/ssh-client-commands/ssh-client-version/) (I have only FreeBSD laptop and server to check) it seems that perhaps there is no reliable version format to parse.
Perhaps there should be local explorer that determines the version using package manager.
Or cdist could try to run ssh with options and if exit status is 255 with specific error (Bad configuration option) then cdist can conclude that those options are not supported.
But then there is a question whether or not run ssh first with all three options (ControlMaster, ControlPersist, ControlPath) and in case of 255 error run it again without ControlPersist because some older versions of ssh support multiplexing but without ControlPersist option.

jdguffey:
@telmich I'm inclined to agree with you. It feels like there should be a more elegant solution but I'd rather have an 80% solution than waiting forever to get an elegant 100% solution. If there's enough value, surely one of us could make a patch to extend the functionality later.

@darko-poljak I like the idea of just giving it a shot and falling back if it doesn't work.

My $0.02.

darko-poljak:
Inspecting options would mean executing something like the following:

```
$ ssh -o ControlPath
command-line line 0: Missing argument.
$ ssh -o ControlMaster
command-line line 0: missing argument.
$ ssh -o ControlPersist
command-line line 0: Missing ControlPersist argument.
$ ssh -o ControlNonexistent
command-line: line 0: Bad configuration option: controlnonexistent
$ ssh -o ControlPersis
command-line: line 0: Bad configuration option: controlpersis
$ ssh -o xxx
command-line: line 0: Bad configuration option: xxx
```
and then checking output for 'Bad configuration option'.

But that would mean 2 or 3 subprocesses (check all three options or only ControlPath and assume ControlMaster and check for ControlPersist) at cdist startup for inspecting ssh options availability.
One env variable and command line option could be defined that would mean to skip inspecting options with values: use these options (my ssh client supports them) or do not use these options (my ssh client does not support them). This will prevent slowdown at each cdist execution.

darko-poljak:
Options inspecting code would be something like the following:

```
wanted_mux_opts = {"ControlPath":"~/.ssh/master-%l-%r@%h:%p", "ControlMaster":"auto", "ControlPersist":"125"}
mux_opts = {}
for x in wanted_mux_opts:
    try:
        subprocess.check_output("ssh -o {}".format(x), stderr=subprocess.STDOUT)
    except subprocess.CalledProcessError as e:
        foo = e.output.decode().lower()
        if not "bad configuration option" in foo:
            mux_opts[x] = wanted_mux_opts[x]
MUX_OPTS = " ".join(["-o {}={}".format(x, mux_opts[x]) for x in mux_opts])
REMOTE_COPY = "scp -o User=root -q " + MUX_OPTS
REMOTE_EXEC = "ssh -o User=root -q " + MUX_OPTS
```

What do you think?

telmich:
I suggest that we give this a try for the master branch and test it for
some days - if it works, we keep it, if it breaks, we fix it.

@asteven?

Darko Poljak [Wed, Mar 09, 2016 at 04:55:21AM -0800]:
> Options inspecting code would be something like the following:
> 
> ```
> wanted_mux_opts = {"ControlPath":"~/.ssh/master-%l-%r@%h:%p", "ControlMaster":"auto", "ControlPersist":"125"}
> mux_opts = {}
> for x in wanted_mux_opts:
>     try:
>         subprocess.check_output("ssh -o {}".format(x), stderr=subprocess.STDOUT)
>     except subprocess.CalledProcessError as e:
>         foo = e.output.decode().lower()
>         if not "bad configuration option" in foo:
>             mux_opts[x] = wanted_mux_opts[x]
> MUX_OPTS = " ".join(["-o {}={}".format(x, mux_opts[x]) for x in mux_opts])
> REMOTE_COPY = "scp -o User=root -q " + MUX_OPTS
> REMOTE_EXEC = "ssh -o User=root -q " + MUX_OPTS
> ```
> 
> What do you think?
> 
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/ungleich/cdist/pull/414#issuecomment-194283737

-- 
Werde Teil des modernen Arbeitens im Glarnerland auf www.digitalglarus.ch!
Lese Neuigkeiten auf Twitter: www.twitter.com/DigitalGlarus
Diskutiere mit auf Facebook:  www.facebook.com/digitalglarus

jdguffey:
@telmich: might it be good to give it a shot in a different branch? It seems like broken code shouldn't really ever be in master. Am I seeing this incorrectly?

asteven:
Nico Schottelius wrote on 03/09/16 17:58:
> I suggest that we give this a try for the master branch and test it for
> some days - if it works, we keep it, if it breaks, we fix it.
>
> @asteven?


So I do a `import cdist` and that alone runs 3 subprocesses?
You gotta be kidding me. If you really want to do this, do it in 
./scripts/cdist

darko-poljak:
@asteven It was just a code snippet showing what would mean inspecting ssh options.
I agree that it is silly  to create several subprocesses just to determine a few options. It seems that there is no easy way to get a list of options. Even showing ssh version and then parsing it includes one subprocess.

Hence, I must agree that it is better to leave such ssh tuning to system admin.
After all, cdist adheres to the KISS principle.

asteven:
Darko Poljak wrote on 03/09/16 21:23:
> @asteven It was just a code snippet showing what would mean inspecting ssh options.
> I agree that it is silly  to create several subprocesses just to determine a few options. It seems that there is no easy way to get a list of options. Even showing ssh version and then parsing it includes one subprocess.
>
> Hence, I must agree that it is better to leave such ssh tuning to system admin.
> After all, cdist adheres to the KISS principle.
>

If you want/need to implement this, do it like this:

- the only place this is ever used is in ./scripts/cdist, so do it there
- if the user has defined the --remote-{copy,exec} arguments just take 
those as they are given
- if the user has defined environment variables CDIST_REMOTE_{COPY,EXEC} 
just take those as they are given
- otherwise check the ssh version or options and based on that construct 
the defaults and use those. Also export these defaults in the environment.
- make sure no subprocesses are executed just by importing a module

telmich:
+1 for Steven's proposal

Steven Armstrong [Thu, Mar 10, 2016 at 03:00:34AM -0800]:
> Darko Poljak wrote on 03/09/16 21:23:
> > @asteven It was just a code snippet showing what would mean inspecting ssh options.
> > I agree that it is silly  to create several subprocesses just to determine a few options. It seems that there is no easy way to get a list of options. Even showing ssh version and then parsing it includes one subprocess.
> >
> > Hence, I must agree that it is better to leave such ssh tuning to system admin.
> > After all, cdist adheres to the KISS principle.
> >
> 
> If you want/need to implement this, do it like this:
> 
> - the only place this is ever used is in ./scripts/cdist, so do it there
> - if the user has defined the --remote-{copy,exec} arguments just take 
> those as they are given
> - if the user has defined environment variables CDIST_REMOTE_{COPY,EXEC} 
> just take those as they are given
> - otherwise check the ssh version or options and based on that construct 
> the defaults and use those. Also export these defaults in the environment.
> - make sure no subprocesses are executed just by importing a module
> 
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/ungleich/cdist/pull/414#issuecomment-194791877

-- 
Werde Teil des modernen Arbeitens im Glarnerland auf www.digitalglarus.ch!
Lese Neuigkeiten auf Twitter: www.twitter.com/DigitalGlarus
Diskutiere mit auf Facebook:  www.facebook.com/digitalglarus

darko-poljak:
I agree with Steven.

darko-poljak:
I added multiplexing options inspection to scripts/cdist.
All it does is that it replaces default cdist.REMOTE_EXEC and cdist.REMOTE_COPY.
I reduced 3 to 2 subprocesses, because ControlMaster and ControlPath are both present from the same version.

Please review.

scripts/cdist
```diff
@@ -20,6 +21,34 @@
 #
 #
 
+def inspect_ssh_mux_opts():
+    import subprocess
+
+    """Inspect whether or not ssh supports multiplexing options"""
+    wanted_mux_opts = {
+        "ControlPath":"~/.ssh/master-%l-%r@%h:%p",
```
telmich: I think the control path should be relative to the temporary cdist directory so the socket vanishes when the run is over

scripts/cdist
```diff
@@ -20,6 +21,34 @@
...
+    for x in check:
+        try:
+            subprocess.check_output("ssh -o {}".format(x),
+                    stderr=subprocess.STDOUT, shell=True)
+        except subprocess.CalledProcessError as e:
+            foo = e.output.decode().lower()
+            if not "bad configuration option" in foo:
+                for o in check[x]:
+                    mux_opts[o] = wanted_mux_opts[o]
+    foo = [" -o {}={}".format(x, mux_opts[x]) for x in mux_opts]
```
telmich: I think foo might not be a great name and I also think we want to have all or none setup, because just some doesn't really help

darko-poljak:
@telmich Yeah, spam would be more pythonic name. Just kidding ;)
You're right, all or none setup. This will also reduce options inspecting code to only 1 subprocess.

darko-poljak:
I will close this pull request.
I will do it in a new fork.
When it's done I will request new pull reqeust.

PR #418, opened 2016-03-18
ssh multiplexing options

darko-poljak:
I re-implemented adding multiplexing options by default if available.
I added two environment variables, CDIST_REMOTE_EXEC and CDIST_REMOTE_COPY.
If not both of them are set then ssh multiplexing options are inspected.
If options are available they are added (all or nothing) to default ssh, scp if environment variables CDIST_REMOTE_EXEC, CDIST_REMOTE_COPY are not set, respectively.
This is done only in scripts/cdist.


asteven:
You are not handling the case where to user called cdist with --remote-copy and/or --remote-exec arguments.

darko-poljak:
@asteven First, default value is constructed. If env variable is set then its value is set as cdist.REMOTE_EXEC/COPY. If env variable is not set then cdist.REMOTE_EXEC/COPY remains its default values and mux opts are appended if available.
Later, argument parser is defined:
`    parser['config'].add_argument('--remote-exec',
         help='Command to use for remote execution (should behave like ssh)',
         action='store', dest='remote_exec',
         default=cdist.REMOTE_EXEC
`
If user supplies command line option then this value is used. If not then default value is used: cdist.REMOTE_EXEC. And this default value is constructed as above.
So, the priority order is: command line options (highest), then environment variable, then default value that is implementation dependent.

asteven:
What I mean is: if the user has defined --remote-copy and/or --remote-exec there's no point/need to spawn a subprocess to figure out defaults that are never used.

darko-poljak:
Ah, yes. Thanks for pointing this out, I missed it. I will fix this.

darko-poljak:
@asteven I made changes to cdist script.
arg parser sets default for remote_copy and remote_exec to env vars if defined else None.
Then after argument parsing if remote_copy or remote_exec is None that means that
user didn't specify option nor env var is defined and implementation default is used with 
inspected mux options.

scripts/cdist
```diff
@@ -101,6 +130,21 @@ def commandline():
...
+        # inspect multiplexing options for default cdist.REMOTE_COPY/EXEC
+        if args_dict['remote_copy'] is None or args_dict['remote_exec'] is None:
+            control_path_dir = tempfile.mkdtemp(prefix="cdist.control.path")
+            import atexit
+            atexit.register(lambda: shutil.rmtree(control_path_dir))
+            mux_opts = inspect_ssh_mux_opts(control_path_dir)
+            if args_dict['remote_exec'] is None:
+                args.remote_exec = cdist.REMOTE_EXEC + mux_opts
+            if args_dict['remote_copy'] is None:
+                args.remote_copy = cdist.REMOTE_COPY + mux_opts
```
telmich: I was wondering - can't we directly save it into the tmp directory that cdist later creates? in that case the mux socket will automatically vanish once cdist is done.

I know that from scripts/cdist the cdist object is (not yet) available, but it sounds a bit more clean to me.

darko-poljak: I'll think about it and see how this can be done.

darko-poljak: You mean on directory that Local class (cdist/exec/local.py) uses for base_path, right?
It uses path from args.out_path. If base_path is not specified then it creates temp directory.
We could do this outside of Local class.
When script/cdist parses args and in the case of "config" command if args.out_path is not specified then script/cdist creates args.out_path temp directory which Config later uses it for Local's base_path and cdist script also uses it for ssh mux socket.
In this way this temp directory is created in the one place and used for both socket and temp files.
What do you think?

darko-poljak:
And here is the change in the code.

scripts/cdist
telmich: Sounds good to me!

darko-poljak: I already made the changes. See comments below and code diff.

telmich:
Looks good to me now - however it conflicts with your own changes now :-)

darko-poljak:
Should have been resolved now.

PR #419, opened 2016-03-22
bugfix for newer zypper (SLES12SP1) releases

dheule:
zypper uses now more fields in output of service list, but url is allways the last one


telmich:
Just a short question for clarification: are you aware that awk is able to search on its own?

echo some text | awk '/some/ { print $1 }'

Not sure if it is suitable to replace the existing grep -E calls, but might be worth a try

dheule:
replacing egrep with awk '/some/ { print $1 }' was not practicable beause /some/ is a regex and we are searching for a url, so i was endig up in a escaping nightmare....
BUT: we dont need regex featrues here, so with a little field separater tweak we can make a string compare with the correct field numbers ... 

Please see the updated code ...

telmich:
Great solution!

PR #421, opened 2016-03-23
fix for __start_on_boot for ALL systemd distros

dheule:
- a new global explorer "init" which retrives the command name of pid 1 (init OR systemd)
- fix __start_on_boot so ALL systemd distros get handled correct out of the box

### DIFF
--- a/cdist/conf/explorer/init
+++ b/cdist/conf/explorer/init
@@ -27,10 +27,8 @@ uname_s="$(uname -s)"
 case "$uname_s" in
     Linux|FreeBSD)
         ps -o comm= -p 1 || true
-        exit 0
     ;;
     *)
-        echo "init exlorer needs to be implemented for $os" >&2
-        exit 1
+        echo "unknown"
     ;;
 esac
--- a/cdist/conf/explorer/init
+++ b/cdist/conf/explorer/init
@@ -18,7 +18,19 @@
 # along with cdist. If not, see <http://www.gnu.org/licenses/>.
 #
 #
-# Check whether the given name will be started on boot or not
+# Returns the process name of pid 1 ( normaly the init system )
+# for example at linux this value is "init" or "systemd" in most cases
 #
 
-ps -o comm= --pid 1
+uname_s="$(uname -s)"
+
+case "$uname_s" in
+    Linux|FreeBSD)
+        ps -o comm= -p 1 || true
+        exit 0
+    ;;
+    *)
+        echo "init exlorer needs to be implemented for $os" >&2
+        exit 1
+    ;;
+esac
--- a/cdist/conf/explorer/init
+++ b/cdist/conf/explorer/init
@@ -29,6 +29,7 @@ case "$uname_s" in
         ps -o comm= -p 1 || true
     ;;
     *)
-        echo "unknown"
+        # return a empty string as unknown value
+        echo ""
     ;;
 esac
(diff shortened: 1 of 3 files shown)


cdist/conf/explorer/init
```diff
@@ -0,0 +1,24 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with cdist. If not, see <http://www.gnu.org/licenses/>.
+#
+#
+# Check whether the given name will be started on boot or not
+#
+
+ps -o comm= --pid 1
```
telmich: Can you quickly verify that this works on "all" OS? And if not, add a || true case?

I'm a bit reluctant to have a Linux only global explorer, even though I clearly see the use case here.

cdist/conf/type/__start_on_boot/explorer/state
```diff
@@ -24,47 +24,51 @@
...
 target_runlevel="$(cat "$__object/parameter/target_runlevel")"
 name="$__object_id"
 
-case "$os" in
-    archlinux)
-        state=$(systemctl is-enabled "$name" >/dev/null 2>&1 \
-            && echo present \
-            || echo absent)
-    ;;
+if [ "$init" == 'systemd' ]; then
```
telmich: I agree with this approach - we should indeed use the "correct" init system, independently of which distro it actually is

telmich:
As suspected, FreeBSD doesn't look compatible:

http://www.freebsd.org/cgi/man.cgi?ps(1)

However it looks rather similar to OpenBSD:

http://man.openbsd.org/OpenBSD-current/man1/ps.1

Maybe embed the ps call in a case $os ... block and leave it open for other os?

p.s.: We might be / are missing an explorer for uname -s - maybe this should be added before and then used for selecting the ps call.

dheule:
what should we return if the os is not implemented in the init explorer ?

- an error --> not implemented os 
- empty string
-  "unknown"

--> I will pulled a version wich returns "unknown", pleae comment ...

telmich:
I think I've a small preference on the empty string, but not a strong opinion. @asteven - what's your 5 Rappen on this one?

cdist/conf/explorer/init
```diff
@@ -0,0 +1,34 @@
...
+#
+#
+# Returns the process name of pid 1 ( normaly the init system )
+# for example at linux this value is "init" or "systemd" in most cases
+#
+
+uname_s="$(uname -s)"
+
+case "$uname_s" in
+    Linux|FreeBSD)
```
telmich: Oh - just seen this line - have you tested that this ps works on FreeBSD? I thought it doesn't!

dheule: Hello Nico,

i have studied the man page, modified the --pid to -p, all other options to
ps should work as stated in the man page.
But: I don't have a FreeBSD to test it ...



2016-03-24 15:57 GMT+01:00 Nico Schottelius <notifications@github.com>:

asteven:
Nico Schottelius wrote on 03/24/16 15:56:
> I think I've a small preference on the empty string, but not a strong opinion. @asteven - what's your 5 Rappen on this one?

Silently don't handle any unknown os. No value results in a empty string.

That's inline with how most other global explorers work, e.g. 
./cdist/conf/explorer/lsb_codename

dheule:
Ok, the return value of unknown case is now empty, if all is ok, please merge or comment ;-)
Thank you

telmich:
Looks good to me! @jdguffey I'd be grateful if you can give this a try on fbsd - to check the ps call

Issue #423, opened 2016-03-30
Remove bashisms


darko-poljak:
Some of types have bashisms.
The most of it is usage of '==' comparison.
E.g. in __pacman_conf, __start_on_boot.
This code won't work on POSIX shell (dash, bourne shell).

darko-poljak:
Merged in https://github.com/ungleich/cdist/commit/7fe4e20f5338cdb44ec25b3cc50e31083d7fa11c

PR #424, opened 2016-03-31
Fix bashisms

darko-poljak:
Hi!
This is related to https://github.com/ungleich/cdist/issues/423.
I ran checkbashisms for cdist/conf/type explorer, gencode-local, gencode-remote and manifest scripts.
This PR is the result of found bashisms removal.
This is necessary if the code should run in posix shell like dash.

The most part of it is because of using '==' instead of '=' comparison operator, suggestions to add shebang and some usage of bash specific parameter expansions like substring expansion.


cdist/conf/type/__pacman_conf/manifest
```diff
@@ -42,11 +42,12 @@ contains_element() {
     MATCH=0
 
     target=$1
-    keys="${@:2}"
+    # keys="${@:2}"
```
telmich: Maybe remove the commented out version? :-)

cdist/conf/type/__pacman_conf/manifest
```diff
@@ -42,11 +42,12 @@ contains_element() {
...
+    shift
 
-
-    for key in ${keys}; do
-        if [ "${key}" == "${target}" ]; then
+    # for key in ${keys}; do
```
telmich: Same here

telmich:
Nice PR - just remove the old cruft and then we're ready for merge!

darko-poljak:
Commented old code is removed.

PR #426, opened 2016-04-01
Fix remote exec bug and bug with save_cache and ssh mux socket file.

darko-poljak:
There is a problem with recent mux options for cdist default ssh/scp usage. ControlPath master socket is created in the temp directory cdist uses for storing files. And this temp dir is in the end saved to cache. But it is an error to move socket file. So I fixed this by changing shutil.move with shutil.copytree, with ignore socket filename pattern, and shutil.rmtree. It is known that Local base_path is a directory and also that destination is a directory. In newer python versions shutil.move also has copy_function parameter but 3.2 don't.
There also is a bug in Remote exec command. I made changes so that cdist does not depend on remote default root shell being posix shell. With this change it can be e.g. csh. But until now when I got ungleich VM for developing rocketchat type I didn't have a real world machine to test with and unit test didn't show this bug.
Without this fix cdist won't work.

### DIFF
--- a/cdist/exec/remote.py
+++ b/cdist/exec/remote.py
@@ -146,10 +146,9 @@ class Remote(object):
         # remotely in e.g. csh and setting up CDIST_REMOTE_SHELL to e.g.
         # /bin/csh will execute this script in the right way.
         if env:
-            cmd.append("/bin/sh")
-            cmd.append("-c")
             remote_env = [" export %s=%s;" % item for item in env.items()]
-            string_cmd = " ".join(remote_env) + " ".join(command)
+            string_cmd = ("/bin/sh -c '" + " ".join(remote_env)
+                + " ".join(command) + "'")
             cmd.append(string_cmd)
         else:
             cmd.extend(command)
--- a/scripts/cdist
+++ b/scripts/cdist
@@ -26,7 +26,8 @@ def inspect_ssh_mux_opts(control_path_dir="~/.ssh/"):
     import subprocess
     import os
 
-    control_path = os.path.join(control_path_dir, "cdist.master-%l-%r@%h:%p")
+    control_path = os.path.join(control_path_dir,
+            "cdist.socket.master-%l-%r@%h:%p")
     wanted_mux_opts = {
         "ControlPath": control_path,
         "ControlMaster": "auto",
@@ -133,24 +134,13 @@ def commandline():
     args_dict = vars(args)
     # if command with remote_copy and remote_exec params
     if 'remote_copy' in args_dict and 'remote_exec' in args_dict:
-        # if out_path is not set then create temp dir here so
-        # Local uses it for base_path and ssh mux socket is
-        # created in it.
-        if args_dict['out_path'] is None:
-            args.out_path = tempfile.mkdtemp()
-            is_temp_dir = True
-        else:
-            is_temp_dir = False
         # if remote-exec and/or remote-copy args are None then user
         # didn't specify command line options nor env vars:
         # inspect multiplexing options for default cdist.REMOTE_COPY/EXEC
         if args_dict['remote_copy'] is None or args_dict['remote_exec'] is None:
-            control_path_dir = args.out_path
-            # only rmtree if it is temp directory;
-            # if user specifies out_path do not remove it
-            if is_temp_dir:
-                import atexit
-                atexit.register(lambda: shutil.rmtree(control_path_dir))
+            control_path_dir = tempfile.mkdtemp()
+            import atexit
+            atexit.register(lambda: shutil.rmtree(control_path_dir))
             mux_opts = inspect_ssh_mux_opts(control_path_dir)
             if args_dict['remote_exec'] is None:
                 args.remote_exec = cdist.REMOTE_EXEC + mux_opts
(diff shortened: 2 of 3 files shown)


cdist/exec/remote.py
```diff
@@ -146,10 +146,9 @@ def run(self, command, env=None, return_output=False):
         # remotely in e.g. csh and setting up CDIST_REMOTE_SHELL to e.g.
         # /bin/csh will execute this script in the right way.
         if env:
-            cmd.append("/bin/sh")
-            cmd.append("-c")
             remote_env = [" export %s=%s;" % item for item in env.items()]
-            string_cmd = " ".join(remote_env) + " ".join(command)
+            string_cmd = ("/bin/sh -c '" + " ".join(remote_env)
+                + " ".join(command) + "'")
```
darko-poljak: The whole command needs to be passed as one argument.

scripts/cdist
```diff
@@ -138,19 +140,11 @@ def commandline():
...
         # if remote-exec and/or remote-copy args are None then user
         # didn't specify command line options nor env vars:
         # inspect multiplexing options for default cdist.REMOTE_COPY/EXEC
         if args_dict['remote_copy'] is None or args_dict['remote_exec'] is None:
             control_path_dir = args.out_path
-            # only rmtree if it is temp directory;
-            # if user specifies out_path do not remove it
-            if is_temp_dir:
-                import atexit
-                atexit.register(lambda: shutil.rmtree(control_path_dir))
```
darko-poljak: atexit rmtree is not needed. Saving cache removes out_path already and this removed code would throw error that path does not exist.

cdist/exec/local.py
```diff
@@ -224,8 +224,16 @@ def save_cache(self):
                 shutil.rmtree(destination)
         except PermissionError as e:
             raise cdist.Error("Cannot delete old cache %s: %s" % (destination, e))
-
-        shutil.move(self.base_path, destination)
```
darko-poljak: Python 3.5 introduced copy_function argument to shutil.move that could be used for this functionality, but Python 3.2 that is minimal python version does not have it so I am using copytree and rmtree.

cdist/exec/local.py
```diff
@@ -224,8 +224,16 @@ def save_cache(self):
...
+        # By default, if available, cdist uses ssh/scp with
+        # multiplexing options. cdist uses user specified out_path
+        # or temp directory for storing files and in the end it
+        # saves this cache to cache directory.
+        # It is an error to move/copy mux ControlPath socket file
+        # to cache so it needs to be ignored when saving to cache.
+        shutil.copytree(self.base_path, destination,
+                ignore=shutil.ignore_patterns(
+                    '*' + cdist.MUX_SOCKET_PREFIX + '*'))
+        shutil.rmtree(self.base_path)
```
darko-poljak: Python 3.5 introduced copy_function argument to shutil.move that could be used for this functionality, but Python 3.2 that is minimal python version does not have it so I am using copytree and rmtree.

telmich:
Hey Darko,

why don't we simply remove the socket when the run is done and keep everything else?

It's not usable in the future anyway :-)

darko-poljak:
As I see it this would mean that in Config.run before calling
`self.local.save_cache()`
we need to call
`os.remove(control_path)`.
Actual file path needs to be resolved as ControlPath is defined as
`cdist.MUX_SOCKET_PREFIX + "%l-%r@%h:%p"`.
We can use
`socket_files = glob.glob(os.path.join(out_path, cdist.MUX_SOCKET_PREFIX + "*"))`
to get the list of files that need to be deleted and then do
`for sockfile in socket_files:
    os.remove(sockfile)`
out_path is user specified output path or temp directory by default.
But then Config class instances needs to be aware of out_path, so this variable need to be added to
Config instances.

This would save us from replacing shutil.move with shutil.copytree and shutil.rmtree.
All other code changes in this PR need to remain.
Should I do it like this?

telmich:
Actually every cache directory exists for only one host (run); so we can actually have a fixed control path name like "ssh_control" instead of using patterns here.

darko-poljak:
Yes, but still Config class instance needs out_path variable.

darko-poljak:
@telmich Or is Local class better place for removing socket?
Perhaps introduce new method cleanup that will be called before save_cache
in Config.run?
cleanup would remove ssh mux socket file.
What do you think?

darko-poljak:
@telmich  I think that the simplest and more clean way would be to separate out_path and ssh mux file path directories. Always put ssh mux control socket file in its own temporary directory and register atexit to remove this temp dir.
Then there would be no need to introduce so much new code just to remove socket file from directory that is to be cached.
There is another such approach in cdist, config.py commandline function creates initial manifest temp file and registers atexit removal of it.
In this way cdist package would remain unaware of mux options (no MUX_SOCKET_PREFIX in __init__.py, no socket removal code in config.py or exec/local.py), as this is only inspected and used from cdist script.

telmich:
Ok - I agree w/ it's own directory to not make code much more complex.

darko-poljak:
I made the changes.
I tested it on real world VM.
It went well.

telmich:
Let's give it a try..

PR #427, opened 2016-04-04
Fix bug with parallel hosts operation and specified output path.

darko-poljak:
When you run cdist with -o and -p options specified and with multiple hosts then cdist fails.
In this way cdist runs parallel, one instance for each host. out_path is specified with -o option, say this is /tmp/foo.
This means that each cdist instance uses this directory for local files.
When one instance is linking emulator e.g. /tmp/foo/bin/__apt_key it succeeds, but when the other tries the same it fails with "file exists".

I propose the following fix.
Local class is changed in such a way that it's base_path constructor argument (out_path) is output parent directory. If argument is None then it creates temp directory as output parent.
Under it Local class creates subdirectory target_host and this directory is then used as Local's base_path.
That means that each instance of parallel execution (each instance of Local class) uses unique output directory which is later cached. After caching output parent directory remains on the system so atexit registered function removes it.

The only change in the code is in cdist/exec/local.py
I tested it with localhost and ungleich VM.
All went well.


cdist/test/exec/local.py
```diff
@@ -35,17 +36,21 @@
 fixtures = op.join(my_dir, 'fixtures')
 conf_dir = op.join(fixtures, "conf")
 
+bin_true = "/usr/bin/true"
+bin_false = "/usr/bin/false"
```
telmich: I'm not a big fan of hardcoding paths for true/false  (like we did before) - maybe use PATH resolution for it?

telmich:
Besides the small true/false issue it looks good to me

darko-poljak:
Ok. I followe the style present. I will fix this test.

darko-poljak:
Fixed.

telmich:
Let's go for 4.0.0!

PR #428, opened 2016-04-06
remove a copy paste artefact which is wrong there

dheule:
I have overseen a cut & paste error which I have found now, sorry


telmich:
Thanks for the follow up!

PR #429, opened 2016-04-12
Adding suuport for OS Devuan

smwltr:
Hi Nico, ungleich, and all cdist contributors.

Here are some changes that will allow cdist to manage Devuan. Devuan is a fork of Debian without systemd. So it's pretty much identical to Debian. However, they have changed the name, and so it was detected as "Unknown OS". Also Devuan is not using systemd. So there will be differences as time goes by.

If there are any problems, please let me know.

Kind regards,

Simon

### DIFF
--- a/cdist/conf/explorer/os
+++ b/cdist/conf/explorer/os
@@ -49,6 +49,11 @@ if [ -f /etc/debian_version ]; then
    echo debian
    exit 0
 fi
+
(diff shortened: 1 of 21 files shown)


telmich:
Thanks a lot and I'm looking forward to more Devuan development!

smwltr:
You are most welcome. I will send my fixes and additions.

PR #430, opened 2016-04-14
Type __pyvenv

darko-poljak:
Hi!
I created new type __pyvenv. It manages python virtual environment using pyvenv for specific user.
It uses python3 venv.
NOTE: This type assumes that pyvenv is already present. The reason to not put something like:
`__package python3-venv`
or
`__package python3`
is that different OS-es and/or OS versions use different pacakges.
E.g. for pyvenv in debian jessie the package 

> python3-venv

 need to be installed.
For arch linux this is 

> python

 package.
For FreeBSD the package is 

> python34

 for python 3.4, or 

> python35

for python 3.5, etc.

This can be assured in manifest init,

I also updated __package_pip: added --runas parameter that specifies user to run pip.
For this this type uses sudo or su, whichever is available. If --runas parameter is not specified
then pip is run without sudo or su (by default, for cdist this means that it is run as root).

The usage of __pyvenv and modified __package_pip is shown in the following example for debina jessie.

```
case "$__target_host" in
    localhost)
        __package python3-venv --state present
        require="__package/python3-venv" __pyvenv /home/darko/testenv --pyvenv "pyvenv-3.4" --owner darko --group darko --mode 740 --state present
        require="__pyvenv/home/darko/testenv" __package_pip docopt --pip /home/darko/testenv/bin/pip --runas darko --state present
    ;;
esac

```

In this example manifest init it is assured that python3-venv packages is installed.
Python virtual environment is created for user darko with specified mode.
Then virtual env's pip is run as user darko to install docopt.

For remaining details see the code.


darko-poljak:
I moved and changed __package_pip local explorers su and sudo to global explorers.
I also added sucmd global explorer.
su, sudo explorers try to find su, sudo and return script code string suitable for running it with two arguments, user and command, respectively. If su, sudo is not found then result string is empty.
sucmd first tries to find sudo, if not found it tries to find su. Result is equivalent to the above.
If the result is assigned to foo then command CMD can be run as user USER with:
`$foo USER CMD`

darko-poljak:
Changed sucmd: prefer su over sudo.

telmich:
Hey Darko! I've had a longer discussion with @asteven and we both agree that even though the idea with the explorers is nice, it is probably superfluous, as every known operating system has "su".

Thus I'd like you to remove the global explorers and rewrite the type to use "su" solely

darko-poljak:
Nico,
I removed global su* explorers and rewrite __package_pip to use "su".

cdist/conf/type/__pyvenv/gencode-remote
```diff
@@ -0,0 +1,67 @@
...
+    present)
+        if [ "$state_should" != "$state_is" ]; then
+            echo $pyvenv $venvparams "$destination"
+        fi
+        if [ \( -n "$owner" -a "$owner_is" != "$owner" \) -o \
+             \( -n "$group" -a "$group_is" != "$group" \) ]; then
+            echo chown -R "${owner}:${group}" "$destination"
+        fi
+        if [ -n "$mode" ]; then
+            echo chmod -R "$mode" "$destination"
```
telmich: I suspect that -R might not be great for chmod - think about 0755 for the whole tree

cdist/conf/type/__pyvenv/man.text
```diff
@@ -0,0 +1,80 @@
+cdist-type__pyvenv(7)
+==================
```
telmich: Attention: the length of = must equal the title length - otherwise asciidoc fails

cdist/conf/type/__pyvenv/gencode-remote
darko-poljak: I implemented this type on the __git basis.
And __git does the same thing with chmod -R.
Should this then be fixed in both, __pyvenv and __git, to only chmod /$__object_id, not
recursively all files inside?

telmich: Ooops. Good catch! In that case I'll merge it as it is to stay consistent and we can think about the problem further in a later release

PR #435, opened 2016-05-22
Option -f: read additional hosts from file (stdin)

darko-poljak:
Hello!

I added -f option to read additional hosts from file or stdin if -f argument is '-'. With this change cdist gets hosts from both, command line arguments and file specified with -f. You can specify both, one of them or none. In case you don't specify hosts on command line or with -f option, cdist reads hosts file from stdin by default. Note that it is an error specifying both initial manifest and host file as '-'.

This is helpful when you need to configure many hosts, e.g. more than 10.
Now you don't need to specify them all at command line, you can have list of them in separate file, each host on its line.

### DIFF
--- a/cdist/conf/type/__block/man.text
+++ b/cdist/conf/type/__block/man.text
@@ -34,7 +34,7 @@ prefix::
    Defaults to #cdist:__block/$__object_id
 
 suffix::
-   the prefix to add after the text.
+   the suffix to add after the text.
    Defaults to #/cdist:__block/$__object_id
 
 state::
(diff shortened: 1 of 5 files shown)


telmich:
Ok for me - @asteven: If it's ok for you, I suggest to make release 4.0.1 with these changes and merge it.

asteven:
Nico Schottelius wrote on 05/23/16 15:56:
> Ok for me - @asteven: If it's ok for you, I suggest to make release 4.0.1 with these changes and merge it.
>

+1

PR #436, opened 2016-05-22
New docs format - reStructuredText (sphinx)

darko-poljak:
Hello!

I have migrated *.text asciidoc format to *.rst reStructuredText format.
I also made changes necessary to build html and man pages using sphinx.
Please review.

Note: build process that is part of man-dist target is tested without git stuff.

To just build html and man pages run:
`make man`

and go to docs/man/_build/html and open index.html for html
or docs/man/_build/man for man pages.


darko-poljak:
Generated html can be seen at:
http://188.226.169.172/cdist-doc/

PR #438, opened 2016-06-03
Bugfix: ssh mux controlpath too long on some envs.

darko-poljak:
Hi!

On some envs ssh multiplexing control path is too long.
For e.g. Linux defines max unis socket path to be 108 bytes.
So I shortened ControlPath cdist sets up by default.

### DIFF
--- a/scripts/cdist
+++ b/scripts/cdist
@@ -26,8 +26,11 @@ def inspect_ssh_mux_opts(control_path_dir="~/.ssh/"):
     import subprocess
     import os
 
-    control_path = os.path.join(control_path_dir,
-            "cdist.socket.master-%l-%r@%h:%p")
+    # socket is always local to each cdist run, it is created in
+    # temp directory that is created at starting cdist, so 
+    # control_path file name can be static, it will always be
+    # unique due to unique temp directory name
+    control_path = os.path.join(control_path_dir, "ssh-control-path")
     wanted_mux_opts = {
         "ControlPath": control_path,
         "ControlMaster": "auto",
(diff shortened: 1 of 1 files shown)


asteven:
I solved this on my cdist by cksumming the target_host.
This way the same control master socket is re-used for connections to the same machine.

e.g.

`
target_host_id="$(echo "${__target_host}" | cksum | awk '{print $1}')"
`

darko-poljak:
To further explain the change in code,
cdist passes ControlPath for default mux options from python code (scripts/cdist),
using fixed part and ssh's placeholders for ControlPath option.
ssh_config man pages recommends using at least %h, %p, and %r placeholders.
%h is replaced with target host name, %p with destination port and %r with remote login username.
cdist will now specify ControlPath to: e.g. **"/tmp/cdist9tmy9n2b/s-%r@%h:%p"**
instead of **"/tmp/tmp9tmy9n2b/cdist.socket.master-%l-%r@%h:%p"** as before where %l is local host name. Using %h, %p and %r ControlPath will be unique for the connection to the same machine.

darko-poljak:
@asteven @telmich do you agree on this changes?

scripts/cdist
```diff
@@ -27,7 +27,7 @@ def inspect_ssh_mux_opts(control_path_dir="~/.ssh/"):
     import os
 
     control_path = os.path.join(control_path_dir,
-            "cdist.socket.master-%l-%r@%h:%p")
+            "s-%r@%h:%p")
```
telmich: Isn't the socket LOCAL to each run anyway? I think we don't even need the variable ssh part in it, because the socket is always within our tempdir for the run and could just use "ssh-control-path" as a static name.

Or am I wrong here?

darko-poljak:
@telmich you are right, at each run cdist creates temp directory (with prefix "cdist") so any path inside that directory will be unique for this run. So I renamed socket to static name "sock".
This will now be for e.g. /tmp/cdist9tmy9n2b/sock.
So its name will not be too long and it will remain unique to cdist run.

telmich:
I suggest to use something more on the line of ssh-control-path or ssh-control-sock or ssh-control-socket - so that somebody else reading the directory content understands for what the socket was thought

darko-poljak:
@telmich I made the change to control path socket name.

PR #439, opened 2016-06-03
Improve error reporting for remote exec.

darko-poljak:
Hello!

When executing cdist and some ssh process error occurs cdist does not report any info about the error.
Just info that command failed.
I made changes to remote.py _run_command so that if ssh/scp (remote_exec/remote_copy) fails cdist will print exit code and output. For reporting output it saves stdout and stderr of called process.
It first removes "-q" argument to command if present so that command won't be quiet.
It does this only on frist run of the remote instance.
Example error that occured as a reason for this change was an error "controlpath too long".
It took around 2 days and adding lines of code to figure out the real cause to ssh command failure. And it was the first ssh command that cdist executes.


darko-poljak:
@asteven @telmich what do you think about this?

telmich:
I think removing a -q is a dangerous idea, as it might be present at different (intended!) locations.

We discussed a long time before that we might want to save stderr/stdout anyway.

So I suggest we remove -q by default, capture the output always, but only present it, if there was an error by default.

We change this behaviour to always display when -v is being used (and clearly also -d)

darko-poljak:
@telmich You are right. So I changed local.py and remote.py to always save stdout and stderr output. In case of -v, -d and/or error this output is reported.
And default REMOTE_EXEC/COPY is changed: no -q flag, so that output is reported.

telmich:
I think we should up the minor version for this feature + create a small blog entry about it

telmich:
Will merge it, when conflicts are resolved

darko-poljak:
@telmich Resolved conflicts.

PR #443, opened 2016-06-07
Implement bash and zsh completions.

darko-poljak:
Hello!

I implemented bash and zsh completions.


telmich:
Looks great - please add changelog request + ping dmitry for updating the debian package to include them

darko-poljak:
@telmich Updated changelog.

Issue #441, opened 2016-06-06
Dependency handling failing


jdguffey:
I have a file, /usr/local/keys/pkg.key, referenced by a manifest as such:

__directory /usr/local/keys --owner root --group wheel --mode 755

require="__directory/usr/local/keys" __file /usr/local/keys/pkg.pub --owner root --group wheel \
   --mode 644 --source $__manifest/global_files/usr/local/keys/pkg.pub

When I execute this manifest, it consistently fails by not creating the __directory/usr/local/keys dependency and so the file has no place to be put.

See the [attached](https://github.com/ungleich/cdist/files/299660/cdist.output.txt) file for debug output (cdist config -dvi)

darko-poljak:
@jdguffey If you put in the manifest only the line:

__directory /usr/local/keys --owner root --group wheel --mode 755

does it create the directory?

Can you send debug output for that case?

jdguffey:
@darko-poljak It does. See the [attached file](https://github.com/ungleich/cdist/files/301403/cdist.output.txt).

The issue I'm having is that cdist isn't even attempting to handle the __directory prior to the __file, even though I'm directing it to with require="..."

darko-poljak:
@jdguffey Hm.. Does this happen for more complex manifest or it happens even for only those two lines:

 __directory /usr/local/keys --owner root --group wheel --mode 755

require="__directory/usr/local/keys" __file /usr/local/keys/pkg.pub --owner root --group wheel \
--mode 644 --source $__manifest/global_files/usr/local/keys/pkg.pub

jdguffey:
It [does](https://github.com/ungleich/cdist/files/301822/cdist.output.txt) work.

# ls -lrt /usr/local/keys/
total 2
-rw-r--r--  1 root  wheel  800 Jun  7 00:18 pkg.pub

jdguffey:
Also, I just discovered something after looking through my hierarchical manifests...

The particular subordinate manifest I was sourcing included a __file/usr/local/keys/pkg.pub, without the __directory/usr/local/keys constraint. Apparently the dependency resolver consistently chose to prioritize that one over the entries in the initial manifest.

This is definitely my error. That said, it does raise an interesting issue. If a single type+object_id combination is used twice, should that be treated as an error? If not, should one win out over the other? If so, which one? The one with more constraints? The one defined first? The one defined last?

darko-poljak:
@jdguffey  From the code,
[...]

PR #445, opened 2016-06-10
Conflicting requirements bugfix.

darko-poljak:
Hello!

This PR is related to the issue https://github.com/ungleich/cdist/issues/441.
With this change cdist now checks if existing and new objects with same parameters have different requirements and if so reports an error.


telmich:
Missing changelog entry, otherwise good

darko-poljak:
@telmich Yeah, I forget it often :)
Updated it.

PR #446, opened 2016-06-11
Feature files export

darko-poljak:
Hello!

Finished feature_files_export branch @telmich started.


telmich:
That's great! We should bounce minor (or even major version) for this feature + write a blog article about it

darko-poljak:
@telmich Resolved conflicts.
I agree, after all the merges I will write blog entries for new things and we can build new version.

PR #451, opened 2016-06-30
Make man pages SEE ALSO by convention. Improve docs building.

darko-poljak:
Hello!

Here are fixes for man pages.
@telmich These are fixes for issues that Dmitry reported.
Man pages should now be by convention, especially SEE ALSO part.
There is also imporvement in building docs.
sphinx target directory is changed to docs/dist so for users who install cdist from git and build docs can now access manpages from docs/dist/man and html documentation from docs/dist/html/index.html.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,8 +20,9 @@ MANIFEST
 dist/
 cdist/version.py
 
-#sphinx build dir
+# sphinx build dirs, cache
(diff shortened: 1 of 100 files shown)


darko-poljak:
@telmich I will now merge this PR and Dmitry can try the changes.

PR #453, opened 2016-06-30
Improve docs build

darko-poljak:
Hello!

This PR is the result of discussion with Dmitry, maintainer of cdist Debian package.
sphinx markup :manpage: is replaced with :strong: in cdist and types' man.rst files so that man page reference in both, man page and html, are in bold.
docs/man is renamed to docs/src, because with new documentation structure it no longer contains man pages source files, but source files (rst) from which both, man and html, are built. 
And there is change in Makefile target.
man target is renamed to docs target.
Some variables in Makefile are renamed accordingly.


telmich:
Great job done here!

PR #454, opened 2016-06-30
Make union of existing and new requirements instead of conflict error.

darko-poljak:
Hello!

Conflicting requirements bugfix caused more errors than benefits.
So here is change that will not raise an error if type object already exists with some different requirements, but will make union of existing requirements set and new requirements set.

### DIFF
--- a/cdist/emulator.py
+++ b/cdist/emulator.py
@@ -201,6 +201,34 @@ class Emulator(object):
             except EnvironmentError as e:
                 raise cdist.Error('Failed to read from stdin: %s' % e)
 
+
+    def record_requirement(self, requirement):
+        """record requirement and return recorded requirement"""
+
+        # Raises an error, if object cannot be created
+        try:
+            cdist_object = self.cdist_object.object_from_name(requirement)
+        except core.cdist_type.NoSuchTypeError as e:
+            self.log.error(("%s requires object %s, but type %s does not"
+                    " exist. Defined at %s"  % (self.cdist_object.name,
+                        requirement, e.name, self.object_source)))
+            raise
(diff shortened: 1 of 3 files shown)


cdist/emulator.py
```diff
@@ -228,40 +256,28 @@ def record_requirements(self):
...
                         self.object_source,
                         reqs))
-                self.log.error(errmsg)
-                raise cdist.Error(errmsg)
+                self.log.debug(dbgmsg)
+                all_reqs = reqs | self._existing_reqs
+                self.log.debug("All requirements: {}".format(all_reqs))
+                for x in all_reqs:
+                    if not x in self.cdist_object.requirements:
+                        self.record_requirement(x)
```
darko-poljak: @telmich 
I have just (yesterday and today morning) looking at that code and thinking about it.
All this changes about dependency conflict and now this change came from this issue report: https://github.com/ungleich/cdist/issues/441.

I have just contacted Jake to send me his manifests that were causing this. And I have forgotten to add you in CC, so I will include you in email.

Because without above code cdist was already making a union of requirements, and this new code has no effect, but for his manifests it seems that it does not, or there was something wrong with his manifests. Without actual manifests I cannot tell if there is something wrong with cdist code or manifests. And when you look at the code there should be everything ok.
So, according to the present facts, we should change this PR to undo commits for fixing above issue. And when we get Jake's manifests then we can see what was wrong.
What do you think?

darko-poljak:
@telmich I have made undo of detecting conflicting requirements.
I left new method record_requirement that is called from record_requirements.
This PR now undoes detecting conflicting requirements, so it will work as before,
continue appending new requirements.
New unit tests show that this works.

PR #456, opened 2016-07-06
Pep8

darko-poljak:
Hello!

I ran pep8 for cdist and fixed code style according to pep8.

### DIFF
--- a/cdist/__init__.py
+++ b/cdist/__init__.py
@@ -44,25 +44,30 @@ BANNER = """
 REMOTE_COPY = "scp -o User=root"
 REMOTE_EXEC = "ssh -o User=root"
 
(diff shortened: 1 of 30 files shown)


telmich:
Omg! Alright, let's get things cleaned up. @asteven you might want to merge again after this merge, so code doesn't diverge too much

asteven:
Guess I will not merge anymore at all. Way to quickly moving target for all my patch sets.

telmich:
@asteven I suggest to move your patch set upstream -- let's talk about it in person next week

telmich:
@asteven I suggest to move your patch sets upstream -- let's talk about it in person next week

PR #457, opened 2016-07-06
Fixes for cdist preos command

toupeira:
I ran into some problems while updating the install image for Debian 8, can you merge these fixes into the 4.0-pre-not-stable branch?


telmich:
Looks great, thanks for the patch!

PR #460, opened 2016-07-07
Fix hostdir

asteven:
This fixes a problem introduced by 3062421.
You can not assume target_host is anything that can be used as a directory name.
E.g. in my setup target_host sometimes looks like this: chroot://hpc-netboot.ethz.ch/local/chroot/chaos-platform-ubuntu-16.04/


cdist/exec/local.py
```diff
@@ -238,7 +239,7 @@ def run_script(self, script, env=None, return_output=False,
                         message_prefix=message_prefix)
 
     def save_cache(self):
-        destination = os.path.join(self.cache_path, self.hostdir)
+        destination = os.path.join(self.cache_path, self.target_host)
```
darko-poljak: @asteven Why self.target_host instead of self.hostdir at line 242? Wouldn't this try to create directory path that contain target_host?

darko-poljak:
@asteven  +1 for hash as directory name.

One suggestion though.

If cdist exits with an error then you have this directory for debug info, but you have to map your target_host to its temp directory.
I know there are debug messages that contain target's temp directory, but I suggest an explicit log.debug line which prints message like "Calculated <target_host> temp directory is <self.hostdir>", for example, after self._hostdir() method call. Or do you think this would be an overhead?

cdist/exec/local.py
asteven: Darko Poljak wrote on 07/07/16 17:52:
>> @@ -238,7 +239,7 @@ def run_script(self, script, env=None, return_output=False,
>>                          message_prefix=message_prefix)
>>
>>      def save_cache(self):
>> -        destination = os.path.join(self.cache_path, self.hostdir)
>> +        destination = os.path.join(self.cache_path, self.target_host)
>
> @asteven Why self.target_host instead of self.hostdir at line 242? Wouldn't this try to create directory path that contain target_host?

In the cache I doen't care. It doesn't hurt anybody there as it's not 
used by any code/shell.

They reason that this all broke for me was that my target_host is 
sometimes a string like 
chroot://some.server.com/path/to/chroot/on/that/server.

The way you implemented it this string 
(chroot://some.server.com/path/to/chroot/on/that/server) ends up in 
CDIST_PATH where the extra colons (:) break the cdist-conf-dirs from 
properly populating which leads to types not being found at runtime.

asteven:
Darko Poljak wrote on 07/07/16 17:58:
> @asteven  +1 for hash as directory name.
>
> One suggestion though.
>
> If cdist exits with an error then you have this directory for debug info, but you have to map your target_host to its temp directory.
> I know there are debug messages that contain target's temp directory, but I suggest an explicit log.debug line which prints message like "Calculated <target_host> temp directory is <self.hostdir>", for example, after self._hostdir() method call. Or do you think this would be an overhead?

As this would printed once per target and run I doen't see any reason 
not to do it.

+1

Good idea!

cdist/exec/local.py
darko-poljak: @asteven Well, it would be more consistent. And if I am right there is a plan to fully implement cache functionality and then I think this will matter.

darko-poljak:
I will merge this PR and make additional changes we agreed on.

PR #462, opened 2016-07-10
Make signed github releases.

darko-poljak:
Hi!

This is the essence for making signed github releases.
signing can be done only for github release.
Bare "git tag" isn't considered github release.
github release has to be created through github web interface or using API.
So it will be tested when making first such release.
The commands and order of them are tested manually with some test repository.


bin/build-helper
```diff
@@ -153,7 +153,81 @@ eof
...
+                      \"target_commitish\":\"master\", \
+                      \"name\": \"${tag}\", \
+                      \"body\":\"${tag}\", \
+                      \"draft\":false, \
+                      \"prerelease\": false}" \
+            "https://api.github.com/repos/ungleich/cdist/releases" || exit 1
+
+        # get release ID
+        repoid=$(curl "https://api.github.com/repos/ungleich/cdist/releases/tags/${tag}" \
+            || python3 -c 'import json; import sys; print(json.loads(sys.stdin.read())["id"])') \
```
telmich: Nice use of python here!

darko-poljak: Thanks

PR #463, opened 2016-07-23
Socket file fix

darko-poljak:
Hi!

With commit [https://github.com/ungleich/cdist/commit/8246642a45563772a7142188d5bca564982d4aaf] that defines fixed ssh ControlPath name there is a bug when running cdsit for more than one host.
ssh resolves ControlPath for current connection.
For the first host file does not exist so it is created. For the next host resolved path is the same path,
so ssh reuses it. This causes that ssh sends commands to the first host through socket file created for the first host, instead for the current host.

This is fixed so that socket file path is constructed such that it is under host's temp directory.
All data directory is moved under "data" sub-directory, this is data that is later cached.
Socket file cannot be cached, i.e., it cannot be moved (removed).
All ssh mux socket code/logic is moved to config.py.
For each host remote_exec/copy is constructed using hosts's temp directory for ControlPath,
if default remote_ commands are used.

@telmich Please, review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -246,3 +246,6 @@ pub:
 
 test:
 	$(helper) $@
+
+pep8:
+	$(helper) $@
--- a/bin/build-helper
+++ b/bin/build-helper
@@ -249,7 +249,7 @@ eof
         # First check everything is sane
         "$0" check-date
         "$0" check-unittest
-        "$0" pep8
+        "$0" check-pep8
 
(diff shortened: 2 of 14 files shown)


darko-poljak:
Note: there are many changes in scripts/cdist file due to pep8.

telmich:
Thanks for the fix!

PR #464, opened 2016-07-26
New types: __keyboard and __ locale_system

darko-poljak:
Hi!

New types from Carlos.
@telmich Please review.

Darko


asteven:
I'd reject the __locale_system type in favour of my __system_locale type which does the same thing but also supports debian, ubuntu, archlinux, redhat and centos. It also supports setting any locale setting instead of just LANG.

@telmich: you have access to it, push it upstream if you want

telmich:
Can you create a separate PR just for __keyboard? I'll add some changes to __locale_system, too, before re-submitting.

PR #465, opened 2016-07-26
New type __hosts

darko-poljak:
Hello!

New type from Dmitry.
@telmich Please review.

Darko


telmich:
Hmm, seeing the type now again, I wonder if it doesn't make a lot more sense to use the IP as the object_id / primary key and add an optional_multiple parameter named "--name"? Even though from the user perspective this might be a bit strange, but it fits more to /etc/hosts style *and* it is better ensured that an IP is only set once

darko-poljak:
Dmitry replies with:
-------------------------------
If $__object_id is ip address, I can easily make
following nonsense:

        __hosts 192.168.1.9 --name foo
        __hosts 10.15.24.56 --name foo

On other hand, if $__object_id is hostname, you can't screw it this
way. $__object_id is something unique, and in /etc/hosts, hostname is,
not ip address.

        __hosts foo --ip 10.12.14.56
        __hosts bar --ip 10.12.14.56

is perfectly valid and results in consistent /etc/hosts.

So, my proposal is apply existing patch.

darko-poljak:
@telmich  ?

darko-poljak:
If an ip address is multiple times in /etc/hosts then first entry with requested hostname is taken.
If a host is defined multiple times then the first entry for that hostname is taken.
Multiple same hostnames in more than one line has no effect, only first entry is ever used.
Multiple same ip addresses with different hostnames are ok, entry for requested hostname is used. 

So, in my opinion, Dmitry is right, hostname in hosts file should be unique and more hostnames can map to the same IP. Hostname should be object id, as is in this PR.

4nd3r:
this type feels really usuful, but using ip address for `$__object_id` sounds more reasonable because `/etc/hosts` uses same order. compromise would be to support both `--ip` and `--name` flags so everyone would be happy then :>

if `--ip` and no `--name` then use `$object_id` as name.
if `--name` and no `--ip` then use `$object_id` as ip.

but of course this will make type more complex and prone to possible bugs :<

telmich:
Would this be conflict free?

How would you solve

__hosts 127.0.0.1 --name localhoRST --state present
__hosts localhoRST --ip 127.0.0.1 --state absent

4nd3r:
> Would this be conflict free?

probably not, haven't tried to write it

> How would you solve

why would you want to do something like that? :>

another thought: what happens if all names from ip are removed?

4nd3r:
okay, I ask and I answer:

```
# First remove all lines for given hostname and then insert again.
# Otherwise, we risk having multiple entries for same hostname.
#
# There is a corner case, which we do not handle. Namely, /etc/hosts
# format allows several host names on single line, like following:
#
#     192.168.15.16 foo bar
#
# If this type manages hostname `foo', then hostname bar will get erased.
```

it's quite common to have entries like `x.x.x.x host.domain.tld host`.
also, this file can be manually edited and all manual edits may get lost when cdist runs over it :<

telmich:
@4nd3r I think it is always ok to lose manual entries, if they conflict with a version control system (not just in this type). However losing settings that *we* made should not be lost...

asteven:
We are managing an entry in the /etc/hosts file. IMHO the proper way to do this is to treat the content of this entry as unknown. In other words, don't depend/care about the values the user wants to set,  use a marker to identify the line we manage instead. Like we do e.g. in __cron.

```shell
__hosts some-object-id \
   --ip 1.2.3.4 \
   --hostname foo \
   --hostname bar
```
Maybe for convenience or to prevent redundancy hostname could default to object_id
```shell
__hosts the.host.name \
   --ip 1.2.3.4
```
Then store those entries in /etc/hosts like this:
```
1.2.3.4 the.host.name #cdist:__hosts/the.host.name
```
Where our entry id created like this #cdist:__hosts/$__object_id.

If the user want's to just remove a entry in /etc/hosts that was put there manually that's a different use case.

darko-poljak:
@asteven @telmich Dmitry sent new patch where he used proposed marker - I have applied it.
I am merging this PR, there is no point in further prattling about it :)

PR #467, opened 2016-07-28
new type __filesystem, first version

dheule:
this type can be used to format luns / disks / partitions if the target has no filesystem.
If it has a filesystem, its only formatet if you add --force.

If a device is mounted, its never atached by this type.

This is the first version, it works (tested) for me on suse SLES and gentoo.

Feel free to add support for more filesystems or os types ....

Thank you,
Daniel


asteven:
Hi Daniel,

Thanks for writing that! I have something similar here, but yours seems to be more complete (I use mine only during install).

As running an explorer is relatively expensive you could implement the following optimization:

```        
man lsblk
...
       -f, --fs
              Output  info  about  filesystems.   This  option  is   equivalent   to
              -o NAME,FSTYPE,LABEL,UUID,MOUNTPOINT.   The  authoritative information
              about filesystems and raids is provided by the blkid(8) command.
...
```

So you could replace all the blkdev_* explorers with one that does:

```
lsblk --fs --list "$blkdev"
```

Then parse the required fields from the table/list where needed which is a simple cheap cut or awk call.

dheule:
Hi Steven,

I think i have found a sexy solution ;-)
Please have a look at the updated explorer and gencode-remote

Thank you

cdist/conf/type/__filesystem/gencode-remote
```diff
@@ -0,0 +1,97 @@
...
+if [ -f "$__object/parameter/force" ]; then
+  # create filesystem even an other filesystem is on disk or the label is not correct, use with caution !
+  forcefs="true"
+else
+  forcefs="false"
+fi
+
+
+
+blkdev_devname="$(grep -P -o2 'NAME="\K[^"]*' "$__object/explorer/lsblk")"
```
asteven: What does the -o2 do? It seems it makes no difference if I run -o or -o2 or -o2678. Not documented in man page either.

asteven:
I would have chosen awk to match on column index in favour of grep. But I'm also ok with this.
Maybe rename the parameter to --device instead of --blkdev as that's also what's used in the `lsblk` man page. But this is just nitpicking.

+1 from me

dheule:
parameter rename to --device done

The -o2 was a leftover from testing with grep which i have overseen, now cleaned up.

asteven:
Maybe could default --device to __object_id as it has to be unique anyway. Would give a slightly more intutitive/shorter signature:

So this:

```
__filesystem dev_sdb --fstype xfs --device /dev/sdb --label Testdisk1
```

could be just this:

```
__filesystem /dev/sdb --fstype xfs --label Testdisk1
```

while still having the option to define with custom __object_id and --device where it makes sense.

Sorry, forgot to mention earlier.

dheule:
Now the --device parameter is optional and defaults to object_id.

I hope the type is now ready for merge,
so if its ok, please merge ;-)

PR #475, opened 2016-08-19
Add target_host file with original host to cache since cache dir is hash

darko-poljak:
@telmich What do you think?
Since target directory in cache is now hash, I think we should add file to host's cache directory with original target_host content.
It will help to find cache dir for particular target host.

It is a little change in save_cache method in exec/local.py.


telmich:
Very good point!

PR #477, opened 2016-08-24
change git.schottelius.org in docs because does not exist

smwltr:


cdist/conf/type/__cdist/man.rst
```diff
@@ -47,7 +47,7 @@ EXAMPLES
     __cdist /home/cdist/cdist
 
     # Use alternative source
-    __cdist --source "git://git.schottelius.org/cdist" /home/cdist/cdist
+    __cdist --source "git://github.com/telmich/cdist" /home/cdist/cdist
```
telmich: It's actually /ungleich/ now instead of telmich - can you update that?

docs/src/cdist-quickstart.rst
```diff
@@ -56,7 +56,7 @@ code into your shell to get started and configure localhost::
     # Get cdist 
     # Mirrors can be found on
     # http://www.nico.schottelius.org/software/cdist/install/#index2h4
-    git clone git://git.schottelius.org/cdist
+    git clone git://github.com/ungleich/cdist
```
telmich: That one here is correct :-)

darko-poljak:
@smwltr Can you update telmich->ungleich in type __cdist  man.rst?
See telmich's comment above.
Then we are ready for PR merge.

Thanks.

darko-poljak:
@smwltr @telmich I will merge this PR and I will update wrong URL.
To update wrong links :)

PR #481, opened 2016-09-02
__package_pkg_openbsd: support --version

andres-erbsen:
This mechanism is used quite heavily on openbsd, for example for choosing between ruby 1.9 and 2.2.


telmich:
Andres, that's amazing, finally we got som emore OpenBSD support!

PR #482, opened 2016-09-12
__user_groups: refactor, support FreeBSD

andres-erbsen:


cdist/conf/type/__user_groups/explorer/oldusermod
```diff
@@ -22,7 +22,8 @@ os="$($__explorer/os)"
 
 if [ "$os" = "netbsd" ]; then
     echo netbsd
-    exit
+elif [ "$os" = "freebsd" ]; then
+    echo freebsd
+else
+    usermod --help | grep -q -- '-A group' && echo true || echo false
 fi
```
telmich: Do we need the os = freebsd / netbsd here at all? If you catch on $os in gencode-remote, it looks like this value is not needed at all

darko-poljak: @andres-erbsen Can you take a look?

darko-poljak:
@andres-erbsen Can you take a look at the review and change/clean explorer accordingly?

darko-poljak:
@telmich I made an explorer cleanup. Any (new) remarks or we should go with a merge?

cdist/conf/type/__user_groups/gencode-remote
```diff
@@ -57,9 +42,25 @@ if [ -z "$changed_groups" ]; then
 fi
 
 for group in $changed_groups; do
-   if [ "$oldusermod" = "true" ]; then
-     echo "usermod $action \"$group\" \"$user\""
+   if [ "$os" = "netbsd" ]; then
```
telmich: So many cases -- can't we just have a case for netbsd and use the old mechanism for all other OS?

darko-poljak:
@telmich FreeBSD does not have usermod at all.
And I believe there is a good reason why old usermode detection is added and new stuff used if usermod is not the old one.

PR #486, opened 2016-10-08
Add warning message for faulty dependencies case.

darko-poljak:
Hi!

I added detection and warning message for case when dependency resolver cannot do it right.

When requirements for the same object are defined in different manifests in init manifest and in some other type manifest and they differs then dependency resolver cannot detect dependencies right. This happens because cdist cannot prepare all objects first and then run objects because some
object can depend on the result of type explorer(s) and explorers are executed during object run. cdist will detect such case and write warning message.

Example for such a case:

    init manifest:
        __a a
        require="__e/e" __b b
        require="__f/f" __c c
        __e e
        __f f
        require="__c/c" __d d
        __g g
        __h h

    type __g manifest:
        require="__c/c __d/d" __a a

cdist will print warning message like this one:

WARNING: cdisttesthost: Object __a/a already exists with requirements:
/usr/home/darko/ungleich/cdist/cdist/test/config/fixtures/manifest/init-deps-resolver /tmp/tmp.cdist.test.ozagkg54/local/759547ff4356de6e3d9e08522b0d0807/data/conf/type/__g/manifest: set()
/tmp/tmp.cdist.test.ozagkg54/local/759547ff4356de6e3d9e08522b0d0807/data/conf/type/__g/manifest: {'__c/c', '__d/d'}
Dependency resolver could not handle dependencies as expected.


telmich:
Looks good to me!

Issue #489, opened 2016-10-30
cdist does not support IPv6 target host address


darko-poljak:
When IPv6 target host address is specified then cdist does not work.
This is because of scp command.
The error is:
cp: cannot create regular file '::1:/var/lib/cdist/conf/explorer': No such file or directory.
This is because ':' is separator for scp command.
IPv6 address should be enclosed between '[' and ']'.
This should be done in cdist python code when constructing remote_copy command.
It should also be done in shell code in types where remote_copy command is used directly like
in __file type gencode-local.

darko-poljak:
See https://github.com/ungleich/cdist/pull/491

darko-poljak:
PR merged.

PR #491, opened 2016-11-02
Support ipv6

darko-poljak:
Hello!

This is related to https://github.com/ungleich/cdist/issues/489.
I fixed this by checking if address is in IPv6 format or not.
If it is then for remote_copy command addr is placed inside [ and ], e.g. [::1].
For remote_ssh this is not necessary.

There are also some types that use remote_copy directly in gencode-local.
In these shell scripts I used very simple method, if address contains colon (:) then I assume
it is IPv6.
I didn't find another method which does not require some tool or is simple enough.

Please review.
And if you have some suggestions for IPv6 check in POSIX shell please tell :)


asteven:
My target_host sometimes looks like this:
ssh+sudo+chroot://host.example.com:1234/path/to/chroot

Your patch would therefore break things for me.

IMHO either document how the user should pass in ipv6 addresses so that scp is happy or, if you want to try to fix this inside of cdist. use the python ipaddress module, try to parse the ip and only escape it if that worked successfully.

asteven:
e.g.

```python
import ipaddress

try:
    ipv6 = ipaddress.IPv6Address(target_host)
    # escape ipv6 address for scp and friends
    target_host = '[%s]' % target_host
except ipaddress.AddressValueError:
    # nothing I could deal with, pass through
    pass
```

asteven:
As for the shell part, maybe it works to re-export the 'escaped' version you handled in the python code?

darko-poljak:
@asteven We cannot instruct users to use '[<IPv6>]' address format because ssh does not work with it. IPv6 address needs to be enclosed with [ and ] only for scp.
So if, for IPv6, we use exported target_host enclosed with [ and ] then
$__remote_exec $__target_host will not work.

I actually use ipaddress module already in python code for remote transfer part.
I am still thinking about the shell code part.
Perhaps we should use some (complex) function that will check for IPv6 address format.

darko-poljak:
@asteven What do you think about checking for IPv6 address using 
`grep -E '^[0-9a-fA-F:]+$'`?
Or `grep -E '^[0-9a-fA-F:.]+$'` which will also detect hybrid dual-stack IPv6/IPv4?

darko-poljak:
@asteven scp actually supports all, IPv4, IPv6 and hostname, to be enclosed in square brackets.
So we could enclose target host in square brackets always.
But this will brake your 'ssh+sudo+chroot://host.example.com:1234/path/to/chroot'.
This is not ssh/scp-like? Is this some custom remote_exec/copy? Should then this custom command handle such enclosing since cdist supports remote_exec/copy to be ssh/scp like?

darko-poljak:
@asteven @telmich I added better check for IPv6 inside shell code.

telmich:
What about ssh? Does it like/accept [...] notation as well? I think I'd prefer consistency for both commands

darko-poljak:
[] notation is only for scp and ssh options with colon as some delimiter.
Ssh with host specified with [] does not work.

Darko

darko-poljak:
@telmich I added IPv6 [ ] notation to docs.
Please review.

telmich:
good for me

PR #494, opened 2016-11-04
__package_upgrade_all shouldn't dist-upgrade by default. also add clean.

4nd3r:
doing blind dist-upgrade can bring surprises :>


4nd3r:
sorry about two different things in one commit.

darko-poljak:
Thanks

PR #495, opened 2016-11-07
add __apt_mark

4nd3r:


cdist/conf/type/__apt_mark/man.rst
```diff
@@ -0,0 +1,47 @@
+cdist-type__apt_mark(7)
+======================
```
telmich: Does it compile with one "=" sign too less?

cdist/conf/type/__apt_mark/gencode-remote
```diff
@@ -0,0 +1,37 @@
...
+    name="$(cat "$__object/parameter/name")"
+else
+    name="$__object_id"
+fi
+
+state="$(cat "$__object/parameter/state")"
+
+case "$state" in
+    auto|manual|hold|unhold)
+        echo "apt-mark $state $name"
```
telmich: Is it actually possible to retrieve the mark before with an explorer?

The notion of cdist is *not* to generate code, if not necessary - so we can eventually check which changes were made by a run.

cdist/conf/type/__apt_mark/man.rst
4nd3r: oups, will fix

cdist/conf/type/__apt_mark/gencode-remote
4nd3r: okay, will fix

4nd3r:
removed "auto" and "manual" since all packages installed with `__package` are set manual anyway and I can't see any reason why one should need to mark something automatically or manually installed with cdist.

another question now is if type should be renamed now? `__apt_mark_hold` maybe? I would leave as is.

also sidenote: `apt-mark (hold|unhold)` is just frondend to `dpkg --(get|set)-selections` and states are stored separately from "auto" and "manual" states.

telmich:
Hey @4nd3r ! From my point of view we can merge the type as it is. If someone finds a use case for auto|manual in the future and cares enough of cdist supporting it, he/she can extend the type.

4nd3r:
ktnx

4nd3r:
@telmich OK, please merge.

4nd3r:
@telmich this type needs 1 or 2 improvements. 

first there should be check if package is installed (if not, then exit) and secondly apt version check. `apt-mark hold` is supproted since wheezy and it's also oldest supported debian release. but life is more difficult and there's probably enough squeezes and lennys in production in closed networks and running that type against squeeze or older will output error.

aboutpackage check. since explorers are run with -e flag... is it okay when explorer just fails? i think not. or where should the package presence check should be? maybe cdist needs separate explorer(s) to check existence of package and its version?

PR #496, opened 2016-11-22
__user type: fix for FreeBSD

AnotherKamila:
On FreeBSD, the `pw useradd` command should have either a different parameter order, or an explicit `-n $username` for the username. Tested on FreeBSD 11.

---
This change is [Reviewable](https://reviewable.io/reviews/ungleich/cdist/496)


telmich:
@jdguffey can you quickly confirm that this still works for you in older fbsd versions?

jdguffey:
I don't have a FreeBSD 9.x box available right now to test, but 10.3 is fine with it. 9 is old enough that it's EoL, anyway, so I don't see any reason for this not to be approved.

Thanks for concern for backwards compatibility, @telmich!

PR #501, opened 2016-11-30
Cache subpath pattern

darko-poljak:
Hello!

I added new option, -C, with which you can specify custom cache subpath pattern.
This pattern is then used to resolve actual cache directory path.
Resolved path is always used as a cache dir subpath.
Pattern supports all placeholders that python datetime.strftime function supports, except %h which
is used as a placeholder for calculated hostdir.
This is current value used and is a default if pattern is not specified or if it is empty.
Additionally %P placeholder is used and it will be substituted with current process id.

I used datetime.strftime function and not time.strftime function because datetime's support %f, microseconds.

Examples pattern resolutions are (in the format PATTERN RESOLVED_VALUE):
<pre>
'' '421aa90e079fa326b6494f812ad13e79'
'/' '421aa90e079fa326b6494f812ad13e79'
'//' '421aa90e079fa326b6494f812ad13e79'
'/%%h' '%h'
'%%h' '%h'
'%P' '3866'
'x%P' 'x3866'
'%h' '421aa90e079fa326b6494f812ad13e79'
'%h/%Y-%m-%d/%H%M%S%f%P' '421aa90e079fa326b6494f812ad13e79/2016-11-30/1804071970013866'
'/%h/%Y-%m-%d/%H%M%S%f%P' '421aa90e079fa326b6494f812ad13e79/2016-11-30/1804071970013866'
'%Y-%m-%d/%H%M%S%f%P/%h' '2016-11-30/1804071970013866/421aa90e079fa326b6494f812ad13e79'
'///%Y-%m-%d/%H%M%S%f%P/%h' '2016-11-30/1804071970013866/421aa90e079fa326b6494f812ad13e79'
'%h/%Y-%m-%d/%H%M%S-%P' '421aa90e079fa326b6494f812ad13e79/2016-11-30/180407-3866'
'%Y-%m-%d/%H%M%S-%P/%h' '2016-11-30/180407-3866/421aa90e079fa326b6494f812ad13e79'
</pre>

If pattern begins with '/', i.e. path separator, then all separators are removed.
If pattern is empty then default hostdir is used.
Default hostdir (currently used) is a hash of a host cdist operates on.

Pattern can also be specified using CDIST_CACHE_PATH_PATTERN env var.

@dheule This PR is a result of your old PR, so please take a look, even try it and send your feedback.
@telmich @asteven What do you think?

Caution.
If, for example, you use '%h/%Y-%m-%d' pattern and then you switch to default value, i.e., '%h',
when saving cache then destination tree is first removed.
This means you will lose '%Y-%m-%d' subdirectories.
**We should discuss on this further.**

### DIFF
--- a/cdist/exec/local.py
+++ b/cdist/exec/local.py
@@ -29,6 +29,9 @@ import subprocess
 import shutil
 import logging
 import tempfile
+import time
+import datetime
+import functools
 
 import cdist
 import cdist.message
@@ -51,7 +54,8 @@ class Local(object):
                  host_dir_name,
                  exec_path=sys.argv[0],
                  initial_manifest=None,
-                 add_conf_dirs=None):
+                 add_conf_dirs=None,
+                 cache_path_pattern=None):
 
         self.target_host = target_host
         self.hostdir = host_dir_name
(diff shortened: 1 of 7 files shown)


cdist/exec/local.py
```diff
@@ -242,18 +246,57 @@ def run_script(self, script, env=None, return_output=False,
...
+                try:
+                    if os.path.isdir(destentry):
+                        shutil.rmtree(destentry)
+                    elif os.path.exists(destentry):
+                        os.remove(destentry)
+                except (PermissionError, OSError) as e:
+                    raise cdist.Error(
+                            "Cannot delete old cache entry {}: {}".format(
+                                destentry, e))
+                shutil.move(srcentry, destentry)
```
darko-poljak: With this change removal of destination dir entries before moving cache path (source) entries is safe.
It will not remove any directories or files not present in source path.
So if you run cdist with different cache path patterns you will not lost cache data.

darko-poljak:
I changed the way of destination removal prior to moving cache data to destination.
With this change you will not lose cache data.

telmich:
Let's try this out in the beta branch with -b enabled

darko-poljak:
@telmich Ok.

I will close this PR then and merge changes into beta.

dheule:
Sorry for the late comment,

but I'm missing a pattern placeholder for the normal hostname ...
I dislike the %h calculated hostnames  ... 

So please add an additional placeholder for the normal hostname ...

Thank you,
Daniel

darko-poljak:
@telmich @asteven @dheule This branch has been merged into beta.

darko-poljak:
@dheule I will add it to beta branch.

dheule:
Thank you ;-)

Daniel

2016-12-01 16:04 GMT+01:00 Darko Poljak <notifications@github.com>:

darko-poljak:
@dheule Added %N (target host) placeholder in beta branch.

PR #502, opened 2016-11-30
Deprecate -d option and make -v option log level counter

darko-poljak:
-d option is now deprecated.
-v can be used to increase log level. The more times you use it the log level increases more.

### DIFF
--- a/docs/src/man1/cdist.rst
+++ b/docs/src/man1/cdist.rst
@@ -50,7 +50,7 @@ All commands accept the following options:
 
 .. option:: -v, --verbose
 
-    Increase log level, be more verbose. Use it more than onceto increase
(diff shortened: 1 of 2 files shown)


docs/src/man1/cdist.rst
```diff
@@ -46,11 +46,13 @@ All commands accept the following options:
...
 .. option:: -d, --debug
 
-    Set log level to debug
+    Set log level to debug (deprecated, use -vvv instead)
 
 .. option:: -v, --verbose
 
-    Set log level to info, be more verbose
+    Increase log level, be more verbose. Use it more than once to increase
+    log level. The order of levels from the lowest to the greatest are:
```
telmich: highest, not greatest!

darko-poljak:
@telmich Fixed.

PR #503, opened 2016-12-02
Added __docker type from asteven, implemented debian support

greendeath:


greendeath:
@darko-poljak Done

darko-poljak:
@greendeath Two more things: missing line in man.rst after line 33; check files and change 'experimentel' to 'experimental'.

PR #504, opened 2016-12-06
The Good, the Bad and the Ugly

darko-poljak:
Hello!

I have made some refactorings. For now only obvious (ugly->bad) parts:
- simplify some APIs
- renamed enable-beta to beta (from beta branch)
- CDIST_BETA env var (from beta branch)
- extracted argparse code into separate file
- extracted functions into separate util files.

@telmich I propose to merge before (if) we continue to do some code cleanup.


cdist/argparse.py
```diff
@@ -0,0 +1,209 @@
...
+    1: logging.WARNING,
+    2: logging.INFO,
+}
+_verbosity_level = collections.defaultdict(
+    lambda: logging.DEBUG, _verbosity_level)
+
+
+def add_beta_command(cmd):
+    if cmd not in BETA_COMMANDS:
+        BETA_COMMANDS.append(cmd)
```
telmich: Maybe a set() might also be appropriate here?

telmich:
Looks good to me!

cdist/argparse.py
darko-poljak: Good point.

darko-poljak:
@telmich Changed list -> set for beta command and args.
Also fixed some missing vars errors.
I will merge this. Of course, we have to test it thoroughly.

_"Beware of bugs in the above code; I have only proved it correct, not tried it."_
_Donald Knuth_

PR #508, opened 2016-12-11
Created __docker_compose type for cdist

greendeath:
Features:
- Install __docker_compose


darko-poljak:
@greendeath Can you change license to GPLv3+?

darko-poljak:
@greendeath Is it possible to have version parameter instead of hardcoded 1.9.0?
Or is something like 'latest' available?

PR #509, opened 2016-12-21
Bugfixed __docker_compose:

greendeath:
If docker-compose is already running, curl won't override the binary
I add an if the file does not exist before download

### DIFF
--- a/cdist/conf/type/__docker_compose/gencode-remote
+++ b/cdist/conf/type/__docker_compose/gencode-remote
@@ -22,7 +22,6 @@
 version="$(cat "$__object/parameter/version")"
 
 # Download docker-compose file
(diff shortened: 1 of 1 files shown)


darko-poljak:
@greendeath But with this change one cannot update docker-compose with new version.
If you try to overwrite a running binary you will get 'Text file busy'. But you can remove it.
How about making a change that will download it to some temporary file and on success
remove original binary and then copy temp file to desired binary location?

Btw. how about adding 'absent' to this type so one can remove docker-compose?

darko-poljak:
@greendeath Instead of removing before copy you can download it to temp file and on success mv temp docker-compose (mv is done in one atomic operation here).

darko-poljak:
If files are on the same physical device then mv uses rename() for which POSIX requires that it is atomic.

PR #510, opened 2016-12-29
F/docker absent

greendeath:


cdist/conf/type/__docker_compose/manifest
```diff
@@ -19,6 +19,15 @@
...
 # Needed packages
-__docker
-__package curl
+if [ ${state} = "present" ]; then
+    __docker
+    __package curl
+elif [ ${state} = "absent" ]; then
+    __file /usr/local/bin/docker-compose --state absent
+else
+    echo "Unknown state: $state_should" >&2
```
darko-poljak: @greendeath I think state_should is a typo, it should be state.

PR #515, opened 2017-01-28
Better describe -v option

darko-poljak:
@telmich Do you agree with this better -v option description?


darko-poljak:
@telmich We can also add -v option's optional argument with which verbosity level can be set directly (last one wins) `-v [<level>]`.
`$ cdist config -v 2`
would then set levle to INFO, same as -vv

PR #516, opened 2017-01-28
Add optional arg to -v option

darko-poljak:
Added optional argument to -v option. Now one can also set verbosity level directly, for example, -v 3 -v2.
If -v is specified without an argument then every such instance increments verbosity level.
If -v with argument is specified after -v instances then argument value is set.
If -v instance is specified after -v with argument instance then verbosity value is first reset and then it increases with every -v instance.
Although now you cannot specify it like -vv or -vvv.

@telmich What do you think?

### DIFF
--- a/cdist/argparse.py
+++ b/cdist/argparse.py
@@ -4,6 +4,7 @@ import multiprocessing
 import os
 import logging
 import collections
+import functools
 
 
 # set of beta sub-commands
@@ -55,7 +56,7 @@ def check_beta(args_dict):
                     raise cdist.CdistBetaRequired(cmd, arg)
 
 
-def check_positive_int(value):
+def check_int(value, min_value=None, max_value=None):
     import argparse
 
     try:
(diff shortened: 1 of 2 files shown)


asteven:
Hi Darko

Thanks for the effort. But I'm a big fat -1 on this.
Overloading -v like that just seems horrible from a usability point of view.

-v means this, except when it means that, but special case if it's that other thing ...

I propose: keep -v[vvv] as it is.
Create a new argument --log-level to set explicitly which overrides both -v and -d.
Nice, simple and understandable out of the box.

darko-poljak:
@asteven Hm... when I think about simplicity and have in mind McIlroy's criticism about modern Linux:

> Everything was small... and my heart sinks for Linux when I see the size of it. [...] The manual page, which really used to be a manual page, is now a small volume, with a thousand options... We used to sit around in the Unix Room saying, 'What can we throw out? Why is there this option?' It's often because there is some deficiency in the basic design — you didn't really hit the right design point. Instead of adding an option, think about what was forcing you to add that option.

then you are right.

If I think about shortcuts and some projects like nmap that uses this verbosity approach then it is not so uncommon. I was thinking about users that want to set desired verbosity level.
Perhaps I was seduced by the dark side :)

I am about to close this PR unmerged. Let us then think about writing negative code and negative options instead.
For example, we can remove -d/--debug in next release.

PR #520, opened 2017-03-07
fix `__user_groups` for FreeBSD

AnotherKamila:
Tested on FreeBSD 11.

---
This change is [Reviewable](https://reviewable.io/reviews/ungleich/cdist/520)

### DIFF
--- a/cdist/conf/type/__user_groups/gencode-remote
+++ b/cdist/conf/type/__user_groups/gencode-remote
@@ -23,6 +23,25 @@ state_should="$(cat "$__object/parameter/state")"
 oldusermod="$(cat "$__object/explorer/oldusermod")"
 os=$(cat "$__global/explorer/os")
 
+if [ "$os" = "freebsd" ]; then
+	cmd="pw groupmod \"\$group\" \$action \"\$user\""
+	addparam="-m"
+	delparam="-d"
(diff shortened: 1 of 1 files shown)


cdist/conf/type/__user_groups/gencode-remote
```diff
@@ -23,16 +23,37 @@ state_should="$(cat "$__object/parameter/state")"
...
 
+if [ "$os" = "freebsd" ]; then
+	cmd="pw groupmod \"\$group\" \$action \"\$user\""
+	addparam="-m"
+	delparam="-d"
+elif [ "$os" = "netbsd" ]; then
+	# NetBSD does not have a command to remove a user from a group
+	cmd="usermod \$action \"\$group\" \"\$user\""
+	addparam="-G"
+	delparam=";;#"
```
telmich: ;;# ? That looks like closing case statement - is that intentional?

AnotherKamila: That's what it looked like in the last stable version :D I will re-do the fix against master (which apparently changed -- for better!).

AnotherKamila:
Uh, sorry, the change was against the 4.3.2 (last stable) branch, it is already fixed (in a cleaner way) in master. Thanks, closing.

PR #526, opened 2017-05-29
new type: __prometheus_server

AnotherKamila:
Installs and sets up Prometheus (https://prometheus.io/).

---
This change is [Reviewable](https://reviewable.io/reviews/ungleich/cdist/526)


darko-poljak:
@AnotherKamila Can you also write doc/man page for the type?

AnotherKamila:
I will create a new PR, because I also added alertmanager.

PR #527, opened 2017-05-31
Shorten ssh control path.

darko-poljak:
On macos the path is too long due to long default TMP dir.

### DIFF
--- a/cdist/config.py
+++ b/cdist/config.py
@@ -27,6 +27,8 @@ import time
 import itertools
 import tempfile
 import socket
+import atexit
+import shutil
 
 import cdist
 import cdist.hostsource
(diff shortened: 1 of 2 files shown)


cdist/config.py
```diff
@@ -176,8 +177,15 @@ def commandline(cls, args):
                               " ".join(failed_hosts))
 
     @classmethod
-    def _resolve_remote_cmds(cls, args, host_base_path):
-        control_path = os.path.join(host_base_path, "ssh-control-path")
+    def _resolve_ssh_control_path(cls):
+        base_path = tempfile.mkdtemp()
+        control_path = os.path.join(base_path, "s")
```
telmich: Won't this create a new control_path on every call? Shouldn't we store it in the Config object as it is valid for the whole run?

darko-poljak: @telmich This is called when constructing remote_exec/copy commands. And this is done only once for a host. remote_exec/copy commands are constructed from the pattern which contains placeholder for control path (if mux options inspection detected that options are supported).
So control path is 'stored' in remote_exec/copy constructed commands.

telmich:
Darko, can you make a new release? @iredmail this should fix your problem

PR #533, opened 2017-06-27
Feature/install

asteven:
Various fixes to cdist install that i collected over the years.


telmich:
LGTM (is that the correct abreviation for looks good to me?)

PR #534, opened 2017-06-27
need a way to set remote.base_path from the command line

asteven:


telmich:
Looks good to me, besides there is no documentation for it :-)

cdist/argparse.py
```diff
@@ -173,6 +173,10 @@ def get_parsers():
            '-s', '--sequential',
            help='operate on multiple hosts sequentially (default)',
            action='store_false', dest='parallel')
+    parser['config_args'].add_argument(
+           '-r', '--remote-out-dir',
+           help='Directory to save cdist output in on the target host',
```
darko-poljak: @asteven  Please fix above help message ('in on').
I would also put this arg before -s.
Let's try to maintain alphabetical order when possible.

asteven:
@telmich there you have your docs ;-)

PR #535, opened 2017-06-27
Ignore dotfiles in types directory

Feuermurmel:
When discovering types, `CdistType` iterates over all entries in `cdist/conf/type`. If anything places a file or directory there, the process fails:

    $ cdist -v config apu.feuermurmel.ch
    ERROR: apu.feuermurmel.ch: Type '.DS_Store' does not exist at /var/folders/r3/lpx2z49s5hs8mmh3vzrgxd9c0000gs/T/tmp4wkbjab2/805ef570a020944e9becf8a2f70ec03e/data/conf/type/.DS_Store
    ERROR: cdist: Failed to configure the following hosts: apu.feuermurmel.ch

In this case it was macOS' Finder which placed a `.DS_Store` file there while I was navigating the source tree.

This is my suggested fix for this type of problem. Please tell me if there's a better alternative.

### DIFF
--- a/cdist/core/cdist_type.py
+++ b/cdist/core/cdist_type.py
@@ -75,7 +75,9 @@ class CdistType(object):
     @classmethod
     def list_type_names(cls, base_path):
         """Return a list of type names"""
-        return os.listdir(base_path)
+        # Ignore files placed in the types directory by a VCS, package manager,
+        # file manager or the like.
+        return [i for i in os.listdir(base_path) if not i.startswith('.')]
 
     _instances = {}
 


darko-poljak:
Perhaps we should use `glob.glob(os.path.join(base_path, '*'))` instead, which treats filenames beginning with a dot (.) as special cases, and does not return them.

@asteven @telmich What do you think?

telmich:
I was thinking about this some time before when .swp files where popping up... so :+1: for ignoring files/types starting with a .

telmich:
Btw, just realising that you are based in .ch - you are invited to join us in Digital Glarus any time if you are interested in a personal meeting - we even have a meetup at http://www.meetup.com/Digital-Glarus-Business-Technology/

darko-poljak:
@telmich I agree on this, but should we use glob.blog instead?

telmich:
No strong opinion at all

darko-poljak:
@telmich Ok. In the end the effect is the same, even the equivalent `if` in glob function.
If we decide to not return list but to use iterator then we can change it to glob.iblog.

@Feuermurmel I ask you to go through other places in cdist code where os.listdir is used and change it there too. E.g. in cdist_type there is
<pre>
self.__explorers = os.listdir(os.path.join(self.absolute_path,
                                                           "explorer"))
</pre>

and there could be '.*' files too.

darko-poljak:
@Feuermurmel In that case I suggest to define new cdist util function that skips '.*' files from os.listdir and use that new function where needed.

darko-poljak:
@Feuermurmel Hi! Any news on this?

darko-poljak:
Closing in favor of https://github.com/ungleich/cdist/pull/544

PR #537, opened 2017-06-28
Any lang type support

darko-poljak:
This is simple change that allows manifest and gencode-* to be written in any language.
For gencode-* is important that they produce shell code.
Change in cdist/exec/local.py consists of the following.
If script is executable then just execute it (assuming it has a proper shebang).
If it is not executable then assume it is a shell script and execute it with '/bin/sh -e' or CDIST_LOCAL_SHELL.
If it is executable but it does not have a proper shebang cdist does not handle that case, it will try
to execute it because it is executable.

All I did to existing types is `chmod +x` for all manifest and gencode-* and ensure shebang is `#/bin/sh -e`.


telmich:
Variable is missing in the reference manpage ;-)

darko-poljak:
CDIST_LOCAL_SHELL is mentioned in cdist man page and in cdist-reference html page. It's an old one :)

Issue #542, opened 2017-07-04
ssh connection multiplexing race condition


asteven:
Someone elses ticket says more then a thousand words https://github.com/ansible/ansible/issues/16731

In short: ssh connection fails with errror -13 if the ssh client tries to use a master connection just while that is being closed due to timeout.

```
...
debug1: channel 1: free: mux-control, nchannels 2
OpenSSH_6.6.1, OpenSSL 1.0.1e-fips 11 Feb 2013
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 56: Applying options for *
debug1: auto-mux: Trying existing master
debug1: multiplexing control connection
debug1: channel 1: new [mux-control]
debug1: channel 2: new [client-session]
debug1: Sending environment.
debug1: Sending env LANG = en_US.UTF-8
debug1: Sending command: /bin/sh -c ' export __target_host=eu-ms-001-02.hpc-lca.ethz.ch;  export __target_hostname=eu-ms-001-02.hpc-lca.ethz.ch;  export __target_fqdn=eu-ms-001-02.hpc-lca.ethz.ch;  export __explorer=/var/lib/cdist/conf/explorer;/bin/sh -e /var/lib/cdist/conf/explorer/cpu_cores'
debug1: mux_client_request_session: master session id: 2
debug1: client_input_channel_req: channel 2 rtype exit-status reply 0
debug1: client_input_channel_req: channel 2 rtype eow@openssh.com reply 0
debug1: channel 2: free: client-session, nchannels 3
debug1: channel 1: free: mux-control, nchannels 2
VERBOSE: eu-ms-001-02.hpc-lca.ethz.ch: Running initial manifest /tmp/tmpel0dijcy/f6e9674130e19aa6ce21ac3dfc79d272/data/conf/manifest/init
Start: Tue Jul  4 23:29:52 CEST 2017
debug1: ControlPersist timeout expired
debug1: channel 0: free: /tmp/tmp3bv0zoi7/s, nchannels 1
Transferred: sent 115836, received 52088 bytes, in 27.4 seconds
Bytes per second: sent 4231.7, received 1902.9
debug1: Exit status -1
End: Tue Jul  4 23:30:00 CEST 2017
VERBOSE: eu-ms-001-02.hpc-lca.ethz.ch: Preparing object __centos_override_default_yum_repos/
[...]

PR #543, opened 2017-07-06
Fix ssh connection multiplexing race condition #542

darko-poljak:
Increase ControlPersist to 2h.
After host run run ssh mux master exit command.
If custom remote exec/copy is specified then do nothing.

If ssh mux master exit command fails then warning is logged and process is continued.
In that case sysadmin should kill ssh [mux] processes or wait for expiration.
Note that control path is removed by cdist so there is no socket file to connect through.


asteven:
not tested but looks good. +1

darko-poljak:
I tested it with my test host with simple manifest.
I even modified config.py (just for test) by adding non existing ssh option to command so that cleanup cmd fails to see the behavior. Warning was printed, cdist finished successfully, control path was removed and `ps` showed ssh [mux] process.

asteven:
Works for me! I don't see what this could break for anyone, so go ahead and merge it in.

PR #547, opened 2017-07-25
Add -p HOST_MAX argument.

darko-poljak:
When operating hosts in parallel do it in HOST_MAX maximum parallel host processes.
If argument is not specified, by default CPU count is used.

### DIFF
--- a/docs/src/man1/cdist.rst
+++ b/docs/src/man1/cdist.rst
@@ -19,14 +19,14 @@ SYNOPSIS
                  [-i MANIFEST] [-j [JOBS]] [-n] [-o OUT_PATH]
                  [-r REMOTE_OUT_DIR] [--remote-copy REMOTE_COPY]
                  [--remote-exec REMOTE_EXEC] [-I INVENTORY_DIR] [-A] [-a]
-                 [-f HOSTFILE] [-p] [-s] [-t]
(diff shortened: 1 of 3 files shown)


dheule:
Cool, we need this feature, since -p with 200 hosts (all complete parallel) was to heavy for the central system and now we can limit that. 

Looking forward to see this in the next release ;-)

telmich:
@dheule we are having similar problems here now -- great to hear that you are also in a nice bigger area of hosts to configure!

docs/src/man1/cdist.rst
```diff
@@ -153,9 +153,10 @@ Configure/install one or more hosts.
 
 .. option:: -j [JOBS], --jobs [JOBS]
 
-    Specify the maximum number of parallel jobs. Global
-    explorers, object prepare and object run are supported
-    (currently in beta).
+    Operate in parallel in specified maximum number of
+    jobs. Global explorers, object prepare and object run
+    are supported. Without argument CPU count is used by
+    default. Currently in beta.
```
darko-poljak: This is for -j option which specifies maximum jobs for global explorers, object prepare and object run within one host.

-p is not in beta.

So we now have -p HOST_MAX and for each host process -j JOBS parallel jobs.

telmich:
Perfect, sorry for the confusion!

PR #551, opened 2017-08-10
Add helpers for cdist config/install integration.

darko-poljak:
Added helper functions for cdist config/install integration, e.g. with django web.
You can configure a host from python shell using these helpers:
<pre>
>>> import cdist.integration
>>> cdist.integration.configure_hosts_simple('185.203.114.185', '/home/darko/.cdist/manifest/init', verbose=cdist.argparse.VERBOSE_TRACE)
</pre>

You should provide your own manifest file and verbose level (default is INFO).

### DIFF
--- a/cdist/config.py
+++ b/cdist/config.py
@@ -308,7 +308,8 @@ class Config(object):
                 initial_manifest=args.manifest,
                 add_conf_dirs=args.conf_dir,
                 cache_path_pattern=args.cache_path_pattern,
-                quiet_mode=args.quiet)
+                quiet_mode=args.quiet,
+                exec_path=sys.argv[0])
 
             remote = cdist.exec.remote.Remote(
                 target_host=target_host,
--- a/cdist/config.py
+++ b/cdist/config.py
@@ -299,6 +299,7 @@ class Config(object):
            so that main process knows for which host function was successful.
         """
 
+        print(args.cache_path_pattern)
         log = logging.getLogger(host)
 
         try:
--- a/cdist/config.py
+++ b/cdist/config.py
@@ -285,6 +285,7 @@ class Config(object):
         """Configure ONE system.
            If operating in parallel then return tuple (host, True|False, )
            so that main process knows for which host function was successful.
+           If operating in sequential mode then cls instance is returned.
         """
 
         log = logging.getLogger(host)
@@ -335,6 +336,8 @@ class Config(object):
 
         if parallel:
             return (host, True, )
+        else:
+            return c
 
     @staticmethod
     def create_base_root_path(out_path=None):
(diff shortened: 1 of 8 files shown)


cdist/config.py
```diff
@@ -308,7 +308,8 @@ def onehost(cls, host, host_tags, host_base_path, host_dir_name, args,
                 initial_manifest=args.manifest,
                 add_conf_dirs=args.conf_dir,
                 cache_path_pattern=args.cache_path_pattern,
-                quiet_mode=args.quiet)
+                quiet_mode=args.quiet,
+                exec_path=sys.argv[0])
```
darko-poljak: Without this one, after changing sys.argv[0] in integration.py in exec/local.py default sys.argv[0] was not seen,
it ended in '' and wrong emulator linking.

cdist/integration.py
```diff
@@ -0,0 +1,89 @@
...
+        hosts = [host, ]
+    elif isinstance(host, collections.Iterable):
+        hosts = host
+    else:
+        raise cdist.Error('Invalid host argument: {}'.format(host))
+
+    # Setup sys.argv[0] since cdist relies on command line invocation.
+    mydir = os.path.dirname(__file__)
+    cdist_bin = os.path.abspath(os.path.join(mydir, '..', 'scripts', 'cdist'))
+    sys.argv[0] = cdist_bin
```
telmich: Wouldn't it be easier to just use "cdist" as sys.argv[0]? I don't see any side effects at the moment

telmich:
Very good! Just one more thing: Maybe mention in the documenetation that the API is not yet stable

cdist/integration.py
darko-poljak: Because in exec/local.py self.exec_path is set to sys.argv[0] and self.exec_path is used for emulator links:
<pre>
    def _link_types_for_emulator(self):
        """Link emulator to types"""
        src = os.path.abspath(self.exec_path)
        for cdist_type in core.CdistType.list_types(self.type_path):
            dst = os.path.join(self.bin_path, cdist_type.name)
            self.log.trace("Linking emulator: %s to %s", src, dst)

            try:
                os.symlink(src, dst)
            except OSError as e:
                raise cdist.Error(
                        "Linking emulator from %s to %s failed: %s" % (
                            src, dst, e.__str__()))
</pre>

darko-poljak:
@telmich Good point, it should be mentioned that API is not stable.

darko-poljak:
@telmich Added warning about unstable helpers.
Any other remark or suggestion?

darko-poljak:
@telmich Merging to master, it will not affect cdist tool, except with new chapter in docs.

PR #554, opened 2017-08-14
Fix remote cmds cleanup

uqam-fob:
When --remote-copy and --remote-exec are provided,
args.remote_cmds_cleanup_pattern is not set.

This patches fixes the evaluation of args.remote_cmds_cleanup_pattern
and prevents cdist from throwing an exception when --remote-copy
and --remote-exec are used.

### DIFF
--- a/cdist/config.py
+++ b/cdist/config.py
@@ -80,6 +80,7 @@ class Config(object):
     @staticmethod
     def construct_remote_exec_copy_patterns(args):
         # default remote cmd patterns
+        args.remote_cmds_cleanup_pattern = ""
         args.remote_exec_pattern = None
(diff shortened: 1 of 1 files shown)


darko-poljak:
@uqam-fob Thanks. But I think it would be better to add
<pre>
args.remote_cmds_cleanup_pattern = ""
</pre>
in the beginning of
<pre>
def construct_remote_exec_copy_patterns(args):
</pre>
allong with 
<pre>
83         args.remote_exec_pattern = None                                          
84         args.remote_copy_pattern = None 
</pre>

uqam-fob:
@darko-poljak Correct. I did not see those 2 lines.

PR #555, opened 2017-08-19
explorer/os: get ID from /etc/os-release

uqam-fob:
/etc/os-release was introduced by systemd[1] and is now
more and more common; even on systems without systemd (e.g. lede).
In addition to detecting the OS based on specific attributes,
this file provides the ID marker to describe the OS.

This commit adds support for OS detection via /etc/os-release.
According to [2], it is already lowercase.

[1] http://0pointer.de/blog/projects/os-release
[2] https://www.freedesktop.org/software/systemd/man/os-release.html

### DIFF
--- a/cdist/conf/explorer/os
+++ b/cdist/conf/explorer/os
@@ -141,10 +141,10 @@ case "$uname_s" in
 esac
 
 if [ -f /etc/os-release ]; then
-	# already lowercase, according to:
-	# https://www.freedesktop.org/software/systemd/man/os-release.html
-	grep '^ID=' /etc/os-release | sed -e 's|^ID=||'
-	exit 0
+   # already lowercase, according to:
+   # https://www.freedesktop.org/software/systemd/man/os-release.html
+   grep '^ID=' /etc/os-release | sed -e 's|^ID=||'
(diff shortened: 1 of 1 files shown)


cdist/conf/explorer/os
```diff
@@ -141,10 +141,10 @@ case "$uname_s" in
 esac
 
 if [ -f /etc/os-release ]; then
-	# already lowercase, according to:
-	# https://www.freedesktop.org/software/systemd/man/os-release.html
-	grep '^ID=' /etc/os-release | sed -e 's|^ID=||'
-	exit 0
+   # already lowercase, according to:
+   # https://www.freedesktop.org/software/systemd/man/os-release.html
+   grep '^ID=' /etc/os-release | sed -e 's|^ID=||'
```
telmich: grep + sed = awk (might seriously read better in awk)

uqam-fob: agreed..

telmich:
Thanks a lot, can be merged from my side - @darko-poljak can you update the change log?

darko-poljak:
@telmich ack

PR #559, opened 2017-08-22
Configfile

darko-poljak:
This is a proposal for cdist config file implementation.

cdist obtains configuration data from the following sources in the following order:

- command-line options
- configuration file specified in command-line (new -g CONFIG_FILE option)
- environment variables
- user's configuration file (~/.cdist/cdist.cfg)
- system-wide configuration file (/etc/cdist.cfg).

cdist configuration file is in the INI file format. Currently it supports only [GLOBAL] section. The possible keywords and their meanings are as follows:

**archiving**
    Use specified archiving. Valid values include: 'none', 'tar', 'tgz', 'tbz2' and 'txz'.
**beta**
    Enable beta functionality. It recognizes boolean values from 'yes'/'no', 'on'/'off', 'true'/'false' and '1'/'0'
**cache_path_pattern**
    Specify cache path pattern.
**conf_dir**
    Comma separated list of configuration directories.
**init_manifest**
    Specify default initial manifest.
**inventory_dir**
    Specify inventory directory.
**jobs**
    Specify number of jobs for parallel processing. If -1 then the default, number of CPU's in the system is used. If 0 then parallel processing in jobs is disabled. If set to positive number then specified maximum number of processes will be used.
**local_shell**
    Shell command used for local execution.
**out_path**
    Directory to save cdist output in.
**parallel**
    Process hosts in parallel. If -1 then the default, number of CPU's in the system is used. If 0 then parallel processing of hosts is disabled. If set to positive number then specified maximum number of processes will be used.
**remote_copy**
    Command to use for remote copy (should behave like scp).
**remote_exec**
    Command to use for remote execution (should behave like ssh).
**remote_out_path**
    Directory to save cdist output in on the target host.
**remote_shell**
    Shell command at remote host used for remote execution.
**verbosity**
    Set verbosity level. Valid values are: 'ERROR', 'WARNING', 'INFO', 'VERBOSE', 'DEBUG', 'TRACE' and 'QUIET'.

### DIFF
--- a/cdist/argparse.py
+++ b/cdist/argparse.py
@@ -3,6 +3,7 @@ import cdist
 import multiprocessing
 import logging
 import collections
+import cdist.configuration
 
 
 # set of beta sub-commands
--- a/cdist/argparse.py
+++ b/cdist/argparse.py
@@ -27,6 +27,17 @@ _verbosity_level = {
     3: logging.DEBUG,
     4: logging.TRACE,
 }
+
+
+# Generate verbosity level constants:
+# VERBOSE_OFF, VERBOSE_ERROR, VERBOSE_WARNING, VERBOSE_INFO, VERBOSE_VERBOSE,
+# VERBOSE_DEBUG, VERBOSE_TRACE.
+this_globals = globals()
+for level in _verbosity_level:
+    const = 'VERBOSE_' + logging.getLevelName(_verbosity_level[level])
+    this_globals[const] = level
+
+
 # All verbosity levels above 4 are TRACE.
 _verbosity_level = collections.defaultdict(
     lambda: logging.TRACE, _verbosity_level)
@@ -87,7 +98,7 @@ def get_parsers():
     parser['loglevel'] = argparse.ArgumentParser(add_help=False)
     parser['loglevel'].add_argument(
             '-q', '--quiet',
-            help='Quiet mode: disables logging, including WARNING and ERROR',
+            help='Quiet mode: disables logging, including WARNING and ERROR.',
             action='store_true', default=False)
     parser['loglevel'].add_argument(
             '-v', '--verbose',
@@ -102,16 +113,15 @@ def get_parsers():
     parser['beta'] = argparse.ArgumentParser(add_help=False)
     parser['beta'].add_argument(
            '-b', '--beta',
-           help=('Enable beta functionality. '
-                 'Can also be enabled using CDIST_BETA env var.'),
+           help=('Enable beta functionality. '),
            action='store_true', dest='beta',
-           default='CDIST_BETA' in os.environ)
+           default=False)
 
     # Main subcommand parser
     parser['main'] = argparse.ArgumentParser(
             description='cdist ' + cdist.VERSION, parents=[parser['loglevel']])
     parser['main'].add_argument(
-            '-V', '--version', help='Show version', action='version',
+            '-V', '--version', help='Show version.', action='version',
             version='%(prog)s ' + cdist.VERSION)
     parser['sub'] = parser['main'].add_subparsers(
             title="Commands", dest="command")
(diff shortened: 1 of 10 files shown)


asteven:
Thanks for setting up the PR.
Didn't look through all the code yet, but I absolutely dont want the config file to go into ~/.cdist.
It should belong to the user, not be part of the conf_dir.
I always use ~/.config/$project/config.ini for these kind of things.
That would be ~/.config/cdist/config.ini in this case.

darko-poljak:
@asteven So you want to follow XDG standard.
Isn't ~/.cdist user's configuration directory?
With $XDG_CONFIG_HOME/cdist/ there would be config directory with user types, etc. and config directory with cdist exec config.
Why not then putting it all together?

darko-poljak:
As a note for the code, I think I will change some implementation details. This is only PoC.

telmich:
I half agree with @asteven, because I see that configurations might be local to a project (i.e. number of parallel jobs or number of parallel hosts).

However I also see that a "global" configuration ala .gitconfig is helpful.

For that reason I suggest the following:

* Look for ~/.cdistconfig (or xdg dir, but I would strongly prefer a regular dotfile) -> apply configuration
* Look for $configdir/.cdistconfig (this is then consistent with our dotfile) -> overwrite parameters from there
* Parse command line / read environment (not sure about the order there) -> overwrite again

telmich:
Btw, I love to discuss with you two :-)

darko-poljak:
@telmich Do I understand?
In "Look for $configdir/.cdistconfig (this is then consistent with our dotfile) -> overwrite parameters from there" you propose new env var that holds directory containing config file?
So it can be set for various projects with specific config values (also can be done with new -g CONFIG_FILE command line option)?

darko-poljak:
@telmich @asteven One question. Assume there is global config file and local config file.
In global config file there are some options defined and in local config file there are some other options defined. And options in local config file replace values in global config file if defined in both.
Should this also be the case with conf_dir or should values for conf_dir accumulate.
E.g. in global config: conf_dir=/dir1:/dir2, in local config: conf_dir=/dir3:/dir4.
Should resulting value be: conf_dir=/dir1:/dir2:/dir3:/dir4 or conf_dir=/dir3:/dir4?
If you look into cdist code, if -c CONF_DIR is specified then this directory is appended to the list of conf dirs.

darko-poljak:
@telmich @asteven I changed the inner implementation.
Previous question stays. Currently, conf_dir values are accumulated.

asteven:
I'm using XDG_CONFIG_HOME because it is used by click and I didn't bother inventing anything else.
http://click.pocoo.org/5/utils/#finding-application-folders

darko-poljak:
@asteven What is your opinion on conf_dir option question above? To accumulate through config files, env var and command line options or not?

darko-poljak:
@asteven Interesting click utils.py source code for get_app_dir:
<pre>
    if WIN:
        key = roaming and 'APPDATA' or 'LOCALAPPDATA'
        folder = os.environ.get(key)
        if folder is None:
            folder = os.path.expanduser('~')
        return os.path.join(folder, app_name)
    if force_posix:
        return os.path.join(os.path.expanduser('~/.' + _posixify(app_name)))
    if sys.platform == 'darwin':
        return os.path.join(os.path.expanduser(
            '~/Library/Application Support'), app_name)
    return os.path.join(
        os.environ.get('XDG_CONFIG_HOME', os.path.expanduser('~/.config')),
_posixify(app_name))
</pre>

So, for POSIX it is ~/.app-name  :)

asteven:
Not sure if I care about a global config file at all.

> 
> E.g. in global config: conf_dir=/dir1:/dir2, in local config: conf_dir=/dir3:/dir4.
> Should resulting value be: conf_dir=/dir1:/dir2:/dir3:/dir4 or conf_dir=/dir3:/dir4?

For me the result would be conf_dir=/dir3:/dir4. It's simple, there out of the box. No complicated code needed to hack on top of configparser to merge values.

> If you look into cdist code, if -c CONF_DIR is specified then this directory is appended to the list of conf dirs.

This makes sense to me. -c should append.

darko-poljak:
@asteven @telmich last commit changes conf_dir accumulation. Only -c command line specified directories are appended to default directories and directories specified in config file or CDIST_PATH.

cdist now obtains configuration data from the following sources in the following
order:

    1. command-line options
    2. configuration file specified at command-line
    3. configuration file specified in CDIST_CONFIG_FILE environment variable
    4. environment variables
    5. user's configuration file (~/.cdist.cfg)
    6. system-wide configuration file (/etc/cdist.cfg).

Command line options win them all.

I hope all of you are happy with system-wide **/etc/cdist.cfg** and user's **~/.cdist.cfg**.

I have chosen **cfg** extension and not **INI** if in the future we change configuration file format, so we don't have to change extension, to be consistent, if it will no longer be in INI format.

asteven:
I absolutely hate dot files in my home directory.

```
[0:33:15] crius:~% ls -ld .* | wc -l
181
0:33:15] crius:~% 
```

Guess that's why I like the XDG stuff. To me it just makes sense, nice and logical
```
~/.config
~/.cache
~/.local
```

But do whatever. It will just be my 182nd dot config file :-(

darko-poljak:
@asteven What about the following for local config file:
check if ~/.cdist.cfg exists and if so use it, otherwise
check if XDG config env var is in environment and if so check for cdist/cdist.cfg under it, otherwise
check for default xdg env var value. 
If none is found then there is no local config file.

asteven:
@darko-poljak That's super nice of you! Special case just for me :-)

But I think it makes things more complicated to have different locations for the same file. It will confuse users and make docs more complicated..
I'll just use and be happy with the file location you choose.
But thanks again ;-)

darko-poljak:
But why not? It is not that complex. And everyone can have local config file where he/she wants.
By default, the first one found from the:
~/.cdist.cfg or get(XDG_CONFIG_HOME, '~/.config')/cdist/cdist.cfg,
additionally it can be specified with CDIST_CONFIG_FILE env var or -g command line option.

PR #560, opened 2017-08-25
Explorer pipes output of apt-key list, which always generates:

mverboom:
Warning: apt-key output should not be parsed (stdout is not a terminal)
on stderr. Redirect stderr of apt-key to /dev/null to prevent output in
cdist run.


telmich:
Looks good to me - can be merged after creating changelog entry - @darko-poljak

PR #562, opened 2017-08-26
__package_pkg_openbsd: support the empty flavor

uqam-fob:
Adds support for specifying an empty flavor by passing --flavor "".

This commit prevents pkg_add(1) from trying to read from stdin for flavor selection
when there is ambiguity and the empty flavor is desired, by specifying --flavor "".


uqam-fob:
This commit prevents pkg_add(1) from trying to read from stdin for flavor selection
when the empty flavor is desired, by specifying --flavor "".

PR #570, opened 2017-09-06
__debug -> __cdist_loglevel

darko-poljak:
@asteven Fixing __debug relict to __cdist_loglevel.


telmich:
nice!

Can be merged from my point of view (w/ changelog obviously ;-)

PR #571, opened 2017-09-07
Make __cdist_loglevel value more expressive.

darko-poljak:
With these changes you no longer have __cdist_loglevel values 40, 30, 10, etc. but "ERROR", "INFO", "DEBUG", etc.
`if [ "$__cdist_loglevel" = "DEBUG" ]`.

### DIFF
--- a/cdist/conf/type/__install_config/gencode-local
+++ b/cdist/conf/type/__install_config/gencode-local
@@ -24,16 +24,16 @@ remote_copy="$__type/files/remote/copy"
 
 cdist_args=""
 case "$__cdist_loglevel" in
-   20)
+   INFO)
       cdist_args="-v"
    ;;
-   15)
+   VERBOSE)
       cdist_args="-vv"
    ;;
-   10)
+   DEBUG)
       cdist_args="-vvv"
    ;;
-   5)
+   TRACE)
       cdist_args="-vvvv"
    ;;
 esac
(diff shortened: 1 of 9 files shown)


cdist/conf/type/__install_config/gencode-local
```diff
@@ -24,16 +24,16 @@ remote_copy="$__type/files/remote/copy"
 
 cdist_args=""
 case "$__cdist_loglevel" in
```
asteven: If cdist itself honors an existing __cdist_loglevel than this is not required anymore.
Nested cdist invocations would automatically inherit this from the parents environment.

darko-poljak: But with current implementation only emulator honors this.
cdist itself, when started, does not.
I just replaced __debug with __cdist_loglevel.

winduptoy:
@asteven @darko-poljak Sorry this is a little bit late after it's already been merged.

What if I wanted to log on all lower levels? For example, if I wanted to log on **INFO** and lower, I would be able to check `$__cdist_loglevel <= 20`. With a string I will only be able to check `$__cdist_loglevel = "INFO" || $__cdist_loglevel = "VERBOSE" || $__cdist_loglevel = "DEBUG" || $__cdist_loglevel = "TRACE"` which is much more difficult to reason about and do incorrectly than with integers.

darko-poljak:
Hm... Then __cdist_log_level will have numerical value again, but we will
alao provide __cdist_log_level_name (INFO, DEBUG, TRACE, etc.).

Darko

PR #572, opened 2017-09-08
Add -l/--log-level option. Honor __cdist_loglevel env var.

darko-poljak:
Added -l LOGLEVEL, --log-level LOGLEVEL opiton for explicitly setting desired log level.
Also, cdist now sets loglevel according to __cdist_loglevel env var if set, according to configuration rules.


asteven:
personally I would use these names for consistency, but don't care to much

--log-level
__cdist_log_level

Otherwise +1

darko-poljak:
@asteven __cdist_loglevel -> __cdist_log_level.
Any other remark?

PR #576, opened 2017-09-12
Log ERROR to stderr and rest to stdout.

darko-poljak:
@asteven As you suggests, this is the PR for separating logging of ERROR to stderr and rest to stdout.


asteven:
Seriously - You are the man!
I think I asked before, but can we hire you?
🥇

Issue #577, opened 2017-09-13
__ssh_authorized_keys overwrites existing keys


pcoder:
If a given key is not a valid public key or if it contains some whitespaces at the end, `__ssh_authorized_keys` seems to overwrite the existing authorized keys.

Here are the two cases that I found erroneous.

1. Invalid public key 
```
__ssh_authorized_keys root \
  --key "testtesttestsetestststestsetest" --state present
```

2. Valid public key with trailing whitespaces
```
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC+sx1TKzHf+yWpMKsB9NE+w1nYxzun1njbzNL4XwmAavcmjTPhRhImD13R/Hg+zyrvPt3/GsNAsbo02QAGhVh7HWduPgAkAUq0Hm3+t0Moi4PmBMRpwZsWq9ot3d6FJUZz5nwx9g/mjsFEg5FqdtMOClcmpSTq14oGovBLjqGA/oCb7cw57hE1S+VLnl0yhF4ppyuvSbTfs9fPdFp2Uk7LjNywcJXXBudbDOerFAPi4X9dE/lH2yJoEyZMsIf9dfoanbROE/kd0wsTNH/+srbI8VSN1xk3FQJNX6EgqvUOBe+lGa7YQV3ifqobOmflUMQ7Fg4vcIiGxc8t/7QjYVel test@archlinux


```

darko-poljak:
@pcoder I will take a look soon.

darko-poljak:
@pcoder Can you give an example?

1. authorized_keys file content before running cdist
2. init manifest content
3. authroized_keys file content after running cdist
4. expected authorized_keys file content after step 3.

So that I know what to reproduce.

pcoder:
@darko-poljak 

1. authorized_keys file content could be containing any number of keys. For instance lets take this, a randomly chosen public key
```
[...]

PR #579, opened 2017-09-14
Bugfix for: __ssh_authorized_keys overwrites existing keys #577

darko-poljak:


darko-poljak:
@pcoder Can you checkout my branch and test this fix, please?

asteven:
+1 if it works, don't have time to test

darko-poljak:
@pcoder Please verify if it works.

pcoder:
@darko-poljak Sorry for the late reply.
I only tested out with limited test cases. And with all of those, the fix works very good. 👍

pcoder:
Just wondering if it would be a good idea to `trim`/`strip` the keys before adding them?

darko-poljak:
@pcoder Hm... yes. And even validate that key is valid. I will add these too.

darko-poljak:
@pcoder It seems that shell code already trims such key. I also added key validation. In case of invalid key error message is printed and type exits with return value 1. Please test.

pcoder:
Hi @darko-poljak, 

__ssh_authorized_keys can be used to configure multiple keys at the same time.
In this regard, I wonder if it would be possible to avoid invalid keys only and configure the other keys that are valid.

Currrently, if I have n keys to configure and if one of them is an invalid one, then none of the keys get configured.

asteven:
Pcoder wrote on 09/16/2017 10:18 AM:
> Hi @darko-poljak,
> 
> __ssh_authorized_keys can be used to configure multiple keys at the same time.
> In this regard, I wonder if it would be possible to avoid invalid keys only and configure the other keys that are valid.
> 
> Currrently, if I have n keys to configure and if one of them is an invalid one, then none of the keys get configured.
> 

Don't use invalid ssh keys? Maybe?

darko-poljak:
I am not sure if even adding parameter - - ignore-invalid is good idea,
user would not know which keys are and which are not configured. Can you
validate keys before calling cdist?

Darko

e-s-a-i:
Checking for a valid key should also follow the correct format, as the file may contain parameters. 
http://man.openbsd.org/OpenBSD-current/man8/sshd.8#AUTHORIZED_KEYS_FILE_FORMAT

pcoder:
Sure, the keys can be validated beforehand

darko-poljak:
@e-s-a-i There are type parametera for options and comment. Key parameter should hold only key type and the key. But we can check if anything after the key, whitespace separated, is present.

asteven:
I've been using this type for +5 years without any problem ever.
If we can easily ignore invalid keys that are already on the target, than let's do that.
But I don't want to jump through hoops and make the code more complicated or fragile for a one time use case that can be solved otherwise. I don't want to start validating input parameters against validity. This will add a maintenance burden for the future. e.g. SSH supports a new type/format/whatever of key -> we have to jump and adapt.
Keep it simple.

darko-poljak:
@asteven I agree, let's keep it simple. The already implemented validation
in this PR makes sense because, for example, entry explorer already works
for valid keys, but not for invalid. So let's check it with the same tr+awk
code. Otherwise, you can add an invalid key but never remove the same. So,
let's stay with this PR's code but don't complicate it further. Agree?

Darko

On Sep 16, 2017 6:57 PM, "Steven Armstrong" <notifications@github.com>
wrote:

I've been using this type for +5 years without any problem ever.
If we can easily ignore invalid keys that are already on the target, than
let's do that.
But I don't want to jump through hoops and make the code more complicated
or fragile for a one time use case that can be solved otherwise. I don't
want to start validating input parameters against validity. This will add a
maintenance burden for the future. e.g. SSH supports a new
type/format/whatever of key -> we have to jump and adapt.
Keep it simple.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/ungleich/cdist/pull/579#issuecomment-329980868>, or mute
the thread

PR #580, opened 2017-09-26
resolved #444 add messaging to __line-type

### LINKED ISSUE #444: __line doesn't seem to log changes
How can I detect in `gencode-remote` if a `__line` type was modified? `$__messages_in` doesn't contain any messages like with `__file` for example.

tom-ee:
This should fix #444 by adding messages for both states (present/absent).
The manpage was updated accordingly.

### DIFF
--- a/cdist/conf/type/__line/gencode-remote
+++ b/cdist/conf/type/__line/gencode-remote
@@ -60,7 +60,7 @@ case "$state_should" in
         # Only replace ' with '"'"' and keep \ as they are
         line_sanitised=$(cat "$__object/parameter/line" | sed -e "s/'/'\"'\"'/g")
         printf '%s' "printf '%s\n' '$line_sanitised' >> $file"
-	echo "line added" >> "$__messages_out"
+        echo "added" >> "$__messages_out"
 
     ;;
     absent)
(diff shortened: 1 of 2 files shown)


cdist/conf/type/__line/gencode-remote
```diff
@@ -60,6 +60,7 @@ case "$state_should" in
         # Only replace ' with '"'"' and keep \ as they are
         line_sanitised=$(cat "$__object/parameter/line" | sed -e "s/'/'\"'\"'/g")
         printf '%s' "printf '%s\n' '$line_sanitised' >> $file"
+	echo "line added" >> "$__messages_out"
```
telmich: The indention looks wrong here

cdist/conf/type/__line/man.rst
```diff
@@ -40,6 +40,13 @@ file
     If supplied, use this as the destination file.
     Otherwise the object_id is used.
 
+MESSAGES
+--------
+line added
+        The line was added.
+
+line removed
+        The line was removed.
```
telmich: I suggest to omit "line", as you will grep for __line as a prefix anyway - just added/removed should be fine

PR #588, opened 2017-10-20
fix explorer/os for devuan ascii

AnotherKamila:
Fix: explorer/os is reporting 'debian' instead of 'devuan' for Devuan Ascii

---
This change is [Reviewable](https://reviewable.io/reviews/ungleich/cdist/588)


telmich:
+ changelog and good to go from my side, @darko-poljak ; I suggest to create a new minor release after this, as it fixes things for all current Devuan users

PR #589, opened 2017-10-23
Add __systemd_unit type

lubo:


4nd3r:
ahh, yes, lets show some love for systemd in cdist :smile:

can we use  `--enabled` (default if state=present) or `--disabled` (default if state=absent) instead of enablement-state?

or just `--state (enabled|disabled)` and add boolean `--remove` (or `--delete`?) for unit removal?

@telmich @darko-poljak ?

lubo:
Since more enablement states can be implemented (see Table 1. in SYSTEMCTL(1) man page), I'd prefer to keep it this way.

If you don't want to have your unit file deleted, just use `__systemd_unit --enablement-state disabled`. I think that when a unit is in absent state, it shouldn't be present on the system at all.

darko-poljak:
@4nd3r @telmich Since I don't use systemd, if you have no remarks on this PR I will merge it.

lubo:
Any update? @4nd3r Any comments on my response?

cdist/conf/type/__systemd_unit/man.rst
```diff
@@ -0,0 +1,84 @@
...
+
+source
+    Path to the config file. If source is '-' (dash), take what was written to
+    stdin as the config file content.
+
+state
+    'present' or 'absent', defaults to 'present' where:
+
+    present
+        the unit is installed, enabled ans started
```
telmich: ans looks like a typo ;-)

telmich:
Re enabling: I suggest prior to this type to merge __service from @steven and maybe keep that functionality in there

cdist/conf/type/__systemd_unit/man.rst
lubo: Fixed.

lubo:
@telmich Wouldn't it make more sense to keep all systemd-specific stuff in __systemd_unit and then just call it from within __service when necessary? That seems like a more modular approach to me.

lubo:
It's been 2 weeks now, is there any particular problem that needs to be solved?

darko-poljak:
@lubo Hi! We decided to merge this PR. If/when general __service type is introduced it will use this type in case of systemd.

PR #590, opened 2017-10-29
Add messaging to __ssh_authorized_key.

tom-ee:
This will provide implicit messaging to __ssh_authorized_keys.

### DIFF
--- a/cdist/conf/type/__ssh_authorized_key/gencode-remote
+++ b/cdist/conf/type/__ssh_authorized_key/gencode-remote
@@ -111,8 +111,10 @@ fi
 case "$state_should" in
    present)
       add_line "$file" "$entry"
+      echo "added to $file ($entry)" >> "$__messages_out"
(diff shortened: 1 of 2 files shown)


cdist/conf/type/__ssh_authorized_key/gencode-remote
```diff
@@ -111,8 +111,10 @@ fi
 case "$state_should" in
    present)
       add_line "$file" "$entry"
+      echo "added to $file ($entry)" >> "$__messages_out"
```
darko-poljak: Why not "add(ed) $entry to $file"?

tom-ee: I started that way around but changed it to ease parsing:
- $entry may or may not have a comment which leads to variable number of fields
- Also checking for $file is "more general" and probably more common than looking for a specific pub-key.

PR #591, opened 2017-10-31
new type: __letsencrypt_cert

AnotherKamila:
Stolen from @telmich , now with FreeBSD support.
Future work: add an `--onchange` parameter to restart services

---
This change is [Reviewable](https://reviewable.io/reviews/ungleich/cdist/591)

### DIFF
--- a/cdist/conf/type/__letsencrypt_cert/gencode-remote
+++ b/cdist/conf/type/__letsencrypt_cert/gencode-remote
@@ -1,0 +1,18 @@
+domain="$__object_id"
+
+exists=$(cat "$__object/explorer/exists")
+webroot="$(cat "$__object/parameter/webroot")"
+admin_email="$(cat "$__object/parameter/admin-email")"
+
+if [ -n "$exists" ]; then
+	exit 0
+fi
+
+cat <<EOF
+if [ ! -d "$webroot" ]; then
+	echo mkdir -p "$webroot"
+fi
+
+certbot certonly -n --agree-tos --email '$admin_email' --quiet --webroot \
+	-w '$webroot' -d '$domain'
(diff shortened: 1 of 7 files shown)


cdist/conf/type/__letsencrypt_cert/gencode-remote
```diff
@@ -0,0 +1,18 @@
+domain="$__object_id"
+
+exists=$(cat "$__object/explorer/exists")
+webroot="$(cat "$__object/parameter/webroot")"
+admin_email="$(cat "$__object/parameter/admin-email")"
+
+if [ -n "$exists" ]; then
```
telmich: I'd usually use [ "$exists" ], but that's more a style question

cdist/conf/type/__letsencrypt_cert/parameter/default/admin-email
```diff
@@ -0,0 +1 @@
+root@localhost
```
telmich: We could put "honeypot@evil-domain.org" here - just kidding, but it would work....

cdist/conf/type/__letsencrypt_cert/gencode-remote
```diff
@@ -0,0 +1,18 @@
...
+	exit 0
+fi
+
+cat <<EOF
+if [ ! -d "$webroot" ]; then
+	echo mkdir -p "$webroot"
```
darko-poljak: @AnotherKamila Shouldn't the above be without echo? It's inside here string.

AnotherKamila: Oh. Um. Yes.

Issue #154, opened 2013-01-15
__line fails, if line contains single/double quotes


telmich:
Single quotes could work, but double quotes cannot. need to find a way to output the line securely without getting it modified

winduptoy:
@telmich I'm having trouble with this one.

On remote machine, `/etc/TEST.txt`
```
Hello
Hello
dc_eximconfig_configtype='local'
Hello
Hello
```


```
echo __line /etc/TEST.txt --line "dc_eximconfig_configtype='local'" --state absent | cdist config --initial-manifest - -v myhost
```
leaves the file unchanged. Is there something I'm missing necessary to escape those single quotes?

darko-poljak:
@winduptoy Can you test a fix from PR: https://github.com/ungleich/cdist/pull/594

winduptoy:
@darko-poljak @telmich #594 works for me.

matthijskooijman:
Seems like this is fixed and can be closed?

asteven:
Just for reference. Issues like this one would also be reliably solved by #663.

PR #594, opened 2017-11-08
Fix for __line containing single quotes: #154.

darko-poljak:
This fixes:
https://github.com/ungleich/cdist/issues/154#issuecomment-342522050

@telmich Take a look.
@winduptoy With this fix you have to use it like:
<pre>
echo __line /etc/TEST.txt --line "dc_eximconfig_configtype=\'local\'" --state absent | cdist config --initial-manifest - -v myhost
</pre>


winduptoy:
@darko-poljak It works! Thank you.

PR #597, opened 2017-11-12
Support explorers when the filesystem is mounted non-executable

uqam-fob:
When the filesystem is mounted with the non-executable option,
explorers are not able to call each others.

This patches allows explorers to use others by invoking them via
`sh`. Note the assumption that explorers are shell scripts.

### DIFF
--- a/cdist/conf/explorer/cpu_cores
+++ b/cdist/conf/explorer/cpu_cores
@@ -22,7 +22,7 @@
 
 # FIXME: other system types (not linux ...)
 
-os=$("$__explorer/os")
+os=$(sh "$__explorer/os")
 case "$os" in
     "macosx")
         echo "$(sysctl -n hw.physicalcpu)"
--- a/cdist/conf/explorer/cpu_sockets
+++ b/cdist/conf/explorer/cpu_sockets
@@ -22,7 +22,7 @@
 
 # FIXME: other system types (not linux ...)
 
-os=$("$__explorer/os")
+os=$(sh "$__explorer/os")
 case "$os" in
     "macosx")
         echo "$(system_profiler SPHardwareDataType | grep "Number of Processors" | awk -F': ' '{print $2}')"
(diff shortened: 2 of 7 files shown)


telmich:
While I do understand the motivation, I am not sure if that is the best approach for multiple reasons:

a) introducing sh (like seen in #598) breaks non-sh explorers (future compat)
b) Adds additional syntax to remember

Is there any space in the filesystem that **does** have exec bits allowed? I guess it would be better to move cdist's tmpdir to that space instead of hard coding sh, which will limit future (planned) extensions of cdist.

uqam-fob:
Agreed, although I cannot find documentation concerning modifying the remote directory where cdist installs its object files. Perhaps I missed something?

darko-poljak:
@uqam-fob You are looking for this one?
https://www.nico.schottelius.org/software/cdist/man/latest/man1/cdist.html#config-install
<pre>
-r REMOTE_OUT_PATH, --remote-out-dir REMOTE_OUT_PATH

    Directory to save cdist output in on the target host.
</pre>

darko-poljak:
@uqam-fob Does the above option solves your problem?

uqam-fob:
I did not try, but no, it does not solve it.

```
$ mount
/dev/wd0a on / type ffs (local, read-only)
/dev/wd0d on /mfs type ffs (local, read-only)
/dev/wd0e on /usr/local type ffs (local, read-only)
mfs:9972 on /dev type mfs (asynchronous, local, noexec, nosuid, size=10240 512-blocks)
mfs:4889 on /etc type mfs (asynchronous, local, nodev, noexec, nosuid, size=20480 512-blocks)
mfs:6175 on /home type mfs (asynchronous, local, nodev, noexec, nosuid, size=10240 512-blocks)
mfs:18409 on /root type mfs (asynchronous, local, nodev, noexec, nosuid, size=10240 512-blocks)
mfs:9485 on /var type mfs (asynchronous, local, noexec, nosuid, size=102400 512-blocks)
```

As you can see, writable filesystems are mounted noexec and those mounted exec are read-only.
You may close this PR. I'll just maintain my own set of explorers.

PR #598, opened 2017-11-12
explorer/memory: Support OpenBSD

uqam-fob:
Adds support to detect the amount of memory available on OpenBSD
systems.

Builds on #597


darko-poljak:
@uqam-fob Can you support OpenBSD without  #597 explorer stuff?

uqam-fob:
See #602 Thanks for the great work.

PR #601, opened 2017-11-13
Merge new type __grub_config into master

Unterstrichmoepunterstrich:
- add a new type for configure GRUB
- type works at Debian, Ubuntu and Devuan
- TODO: add some logic e.g. when running NetBSD then don't care about GRUB_CMDLINE_LINUX, but afaik NetBSD isn't so good supported.


cdist/conf/type/__grub_config/manifest
```diff
@@ -0,0 +1,398 @@
...
+    --line GRUB_TIMEOUT=\""$GRUB_TIMEOUT"\" \
+    --state present
+fi
+
+if [ -n "$GRUB_DISTRIBUTOR" ]; then
+__line GRUB_DISTRIBUTOR --file "$grub_default_path" \
+    --line GRUB_DISTRIBUTOR=\""$GRUB_DISTRIBUTOR"\" \
+    --state present
+fi
+
```
darko-poljak: @Unterstrichmoepunterstrich Since there are many lines of the same form I would go with something like: 
<pre>
for var in GRUB_DEFAULT \
               GRUB_TIMEOUT \
               ....; do
    __line $var --file "$grub_default_path" --line $var=\""$$var...
done
</pre>
And you probably need dependency on __file for each __line, so this would be:
<pre>
for var in GRUB_DEFAULT \
               GRUB_TIMEOUT \
               ....; do
    eval "var_val=\$$var"  # get a value from variable which var is referencing
    require="__file/..."  __line $var --file "$grub_default_path" --line $var=\""${var_val}"\" --state present
done
</pre>
This way you have list of many vars, but only one line using require="__file..." __line...

uqam-fob:
Just a thought: since invoking the type requires knowledge of GRUB's configuration file (one has to know the variables), how is that different from storing the file and calling the update script. Example:

```
__file /etc/default/grub \
   --state present \
    --owner 0 \
    --group 0 \
    --mode 0644 \
    --source "${config}"

__process update-grub \
    --start 'update-grub'
```

Maybe simply wrapping around those would be more appropriate? Is the type going to be updated with every new flags? What if there's a flag that is not supported?

For what it's worth, how about reading the flags from `${__type}/parameter/*` instead of writing them in the manifest? Also, this can be compacted into 1 for-loop (e.g. `foreach (vars) { get-default-or-option; set-line; }`). Also, this creates an unfriendly configuration file on the target. How about creating the file in a single pass? Plus this causes multiple connections to be made; performance-wise, you cannot have less efficient.

cdist/conf/type/__grub_config/manifest
```diff
@@ -0,0 +1,186 @@
...
+[ -f "$__object/parameter/GRUB_OS_PROBER_SKIP_LIST" ] && GRUB_OS_PROBER_SKIP_LIST=$(cat "$__object/parameter/GRUB_OS_PROBER_SKIP_LIST")
+[ -f "$__object/parameter/GRUB_DISABLE_SUBMENU" ] && GRUB_DISABLE_SUBMENU=$(cat "$__object/parameter/GRUB_DISABLE_SUBMENU")
+[ -f "$__object/parameter/GRUB_ENABLE_CRYPTODISK" ] && GRUB_ENABLE_CRYPTODISK=$(cat "$__object/parameter/GRUB_ENABLE_CRYPTODISK")
+[ -f "$__object/parameter/GRUB_PRELOAD_MODULES" ] && GRUB_PRELOAD_MODULES=$(cat "$__object/parameter/GRUB_PRELOAD_MODULES")
+[ -f "$__object/parameter/GRUB_RECORDFAIL_TIMEOUT" ] && GRUB_RECORDFAIL_TIMEOUT=$(cat "$__object/parameter/GRUB_RECORDFAIL_TIMEOUT")
+[ -f "$__object/parameter/GRUB_RECOVERY_TITLE" ] && GRUB_RECOVERY_TITLE=$(cat "$__object/parameter/GRUB_RECOVERY_TITLE")
+[ -f "$__object/parameter/GRUB_HIDDEN_TIMEOUT" ] && GRUB_HIDDEN_TIMEOUT=$(cat "$__object/parameter/GRUB_HIDDEN_TIMEOUT")
+[ -f "$__object/parameter/GRUB_HIDDEN_TIMEOUT_QUIET" ] && GRUB_HIDDEN_TIMEOUT_QUIET=$(cat "$__object/parameter/GRUB_HIDDEN_TIMEOUT_QUIET")
+[ -f "$__object/parameter/GRUB_HIDDEN_TIMEOUT_BUTTON" ] && GRUB_HIDDEN_TIMEOUT_BUTTON=$(cat "$__object/parameter/GRUB_HIDDEN_TIMEOUT_BUTTON")
+
```
darko-poljak: You could also simplify code by removing above lines and put that logic inside loop body below.

darko-poljak:
@telmich @asteven Your thoughts on this type?

cdist/conf/type/__grub_config/manifest
```diff
@@ -0,0 +1,186 @@
...
+#
+
+os=$(cat "$__global/explorer/os")
+os_version=$(cat "$__global/explorer/os_version")
+grub_default_path="/etc/default/grub"
+# check os
+case "$os" in
+    devuan)
+        case "$os_version" in
+            jessie) GRUB_DISTRIBUTOR=`lsb_release -i -s 2> /dev/null || echo Devuan`
```
darko-poljak: Lines like above should be refactored. I propose writing type's explorer that gets lsb_release -i -s from remote,
and here use it something like GRUB_DISTRIBUTOR=$(cat "$__object/explorer/lsb_release"),
without os and os_version logic - this should be handled in explorer.

lubo:
Since I'm a cdist novice, I don't feel qualified to review this code, but I feel like I could use the answer to educate myself. So:

Why in particular is having this type better than just doing something like `__config_file /etc/default/grub --onchange "grub-mkconfig -o /boot/grub/grub.cfg"`?

telmich:
There are some noteworthy things in this PR, some of which I want to point out:

* supporting all or many options is usually not a good start, reduce it to what is actually used
* including OS specific calls without OS checking should be avoided (gencode-remote)
* if there is OS specific code and OS is not supported, the cdist type should error out
* as already pointed out, don't repeat yourself, no code duplications
* PR should not contain code of other types

Overall, the purpose of a type should be to make life easier, to reduce amount of work required at other places and maybe abstract away.

That said, I think it's a good first start from @Unterstrichmoepunterstrich, but the PR needs major rework to be clean. For that reason I suggest that we close this PR and wait for a newly cleaned up version.

darko-poljak:
In waiting clean version we close this PR.

PR #603, opened 2017-11-18
add --minute to __cron

AnotherKamila:
We don't want this to run _every_ minute (default for `--minute` is `*`, so without the fix it would).

---
This change is [Reviewable](https://reviewable.io/reviews/ungleich/cdist/603)

### DIFF
--- a/cdist/conf/type/__letsencrypt_cert/manifest
+++ b/cdist/conf/type/__letsencrypt_cert/manifest
@@ -68,4 +68,5 @@ esac
 __cron letsencrypt-certbot  \
     --user root \
     --command "$certbot_fullpath renew -q" \
-    --hour 0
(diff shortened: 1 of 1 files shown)


telmich:
Good point! But shouldn't it be 42? (not a too serious note though)

darko-poljak:
@AnotherKamila I am curious, why/how did you choose exactly 47?

AnotherKamila:
Because 47 is [the best random number](http://www.47.net/47society/47story/), obviously.

darko-poljak:
@AnotherKamila Since 47 is also safe prime, supersingular prime, Lucas prime, Eisenstein's prime, etc. (https://en.wikipedia.org/wiki/47_(number)) I am now merging this PR :D

PR #604, opened 2017-11-23
merge __grub_config into master

Unterstrichmoepunterstrich:
I cleaned the code. Afaik we don't detect NextBSD in cdist, so the var is possible to use it, but the type fails.
So I would merge it into master and update it for other OS, too.
What do you think @telmich @asteven?


Unterstrichmoepunterstrich:
ping @telmich @asteven

asteven:
I'm -1 on this type and especially its implementation.

First, I personaly have no use for it. Of course it may be usefull for others.
My grub config is created based on input from multiple types. e.g. I might have type that enables the serial console. So it has to add 'console=tty0 console=ttyS0,115200' to the grub command line. Then I have a other types that also have to contribute to the grub command line. So this type is not useful for me.

Second, I dislike the implementation. There is a lot of code redundancy.
It has almost 70 lines of code that do nothing but setting GRUB_DISTRIBUTOR="$lsb_id".
Then it has ~50 lines of code that read all parameters into shell variables just to eval (!) them and then pass that key/value pair to the __line type. This whole type could be rewritten in 20 lines of shell code.

Third. This pull requests also changes __letsencrypt_cert which has nothing to do with grub.

PR #613, opened 2018-01-19
Handle stdin to __systemd_unit correctly

thriqon:
When directly piping in the source of a systemd unit file, cdist
errored out. This is due to not propagating the stdin file to the
underlying __config_file type (which already contains code for this
case, but this has to be duplicated in __systemd_unit).

The following example thus works:

    __systemd_unit test.service --source - <<EOF
    [Service]
    ExecStart=/bin/sleep 3600
    EOF

### DIFF
--- a/cdist/conf/type/__systemd_unit/manifest
+++ b/cdist/conf/type/__systemd_unit/manifest
@@ -34,6 +34,11 @@ if [ -z "${source}" ] && [ "${state}" != "absent" ]; then
     exit 0
 fi
 
+# stdin is not propagated automatically to sub-objects
+if [ "${source}" = "-" ]; then
+    source="${__object}/stdin"
+fi
+
 __config_file "/etc/systemd/system/${name}" \
     --mode 644 \
     --onchange "systemctl daemon-reload" \


cdist/conf/type/__systemd_unit/manifest
```diff
@@ -34,6 +34,11 @@ if [ -z "${source}" ] && [ "${state}" != "absent" ]; then
     exit 0
 fi
 
+# stdin is not propagated automatically to sub-objects
+if [ "x$source" = "x-" ]; then
```
lubo: What is going on with this _x_ thing? Why is a simple `"${source}" = "-"` not enough?

telmich: "old school", but I let somebody else elaborate on it :-)

thriqon:
I changed the code according to your suggestions.

darko-poljak:
@lubo Do you have any other remark or this PR can be merged?

Darko

On Jan 21, 2018 10:53, "Jonas Weber" <notifications@github.com> wrote:

I changed the code according to your suggestions.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/ungleich/cdist/pull/613#issuecomment-359236268>, or mute
the thread

cdist/conf/type/__systemd_unit/manifest
lubo: Uh, I see now (ref: https://stackoverflow.com/a/174288/1928832). But it'd be a bit redundant in this case, really.

lubo:
@darko-poljak No, I've approved these changes already. Thank you.

PR #617, opened 2018-01-30
Maxage for  package update index

tom-ee:
Add an optional `--maxage <seconds>` parameter for apt in __package_update_index.
Also add messaging: generate a message if the index was updated (for apt) along with the "age".

### DIFF
--- a/cdist/conf/type/__package_update_index/explorer/currage
+++ b/cdist/conf/type/__package_update_index/explorer/currage
@@ -1,32 +1,17 @@
-#!/bin/sh
-#
-# 2018 Thomas Eckert (tom at it-eckert.de)
-#
-# This file is part of cdist.
-#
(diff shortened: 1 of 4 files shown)


cdist/conf/type/__package_update_index/explorer/currage
```diff
@@ -0,0 +1,32 @@
...
+
+case "$os" in
+        debian|ubuntu|devuan)
+		if [ -f "/var/cache/apt/pkgcache.bin" ]; then
+			echo $(($(date +"%s")-$(stat --format '%Y' /var/cache/apt/pkgcache.bin)))
+		else
+			echo 0
+		fi
+                ;;
+        *)      ## not supported $os
```
darko-poljak: Please write error message to stderr. You can find examples in other types.

PR #618, opened 2018-01-30
remove `singleton` from `__apt_update_index`

tom-ee:
If multiple apt sources were added by `__apt_source` the call of
`__apt_update_index` emitted a warning (cannot resolve dependencies) due
to conflicting auto-require.

This commit allows `__apt_update_index` to be used multiple times.

A note was added to the man-page that this should not be used directly.

The two types using `__apt_update_index` (`__apt_source` and
`__apt_ppa`) were updated accordingly.

### DIFF
--- a/cdist/conf/type/__apt_ppa/manifest
+++ b/cdist/conf/type/__apt_ppa/manifest
@@ -1,6 +1,7 @@
 #!/bin/sh -e
 #
 # 2011-2016 Steven Armstrong (steven-cdist at armstrong.cc)
+# 2018 Thomas Eckert (tom at it-eckert.de)
 #
 # This file is part of cdist.
 #
@@ -27,4 +28,4 @@ require="__package/software-properties-common" \
    --source "$__type/files/remove-apt-repository" \
    --mode 0755
 
-require="$__object_name" __apt_update_index
+require="$__object_name" __apt_update_index "$__object_id"
--- a/cdist/conf/type/__apt_source/manifest
+++ b/cdist/conf/type/__apt_source/manifest
@@ -1,6 +1,7 @@
 #!/bin/sh -e
 #
 # 2011-2013 Steven Armstrong (steven-cdist at armstrong.cc)
+# 2018 Thomas Eckert (tom at it-eckert.de)
 #
 # This file is part of cdist.
 #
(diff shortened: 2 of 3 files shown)


darko-poljak:
@tom-ee Can you write your case with cdist output? I have used multiple `__apt_source` and with proper `require` there is no problem. Moreover,  `__apt_update_index` is expensive.

tom-ee:
Sure, here we go.

Warnings:
```
# cdist config -v client01.local:
(...)
WARNING: client01.local: Object __apt_update_index/ already exists with requirements:
/tmp/tmpv_r5stjc/68878079ef13f7cb21067fd9e16ae2d1/data/conf/type/__apt_source/manifest /tmp/tmpv_r5stjc/68878079ef13f7cb21067fd9e16ae2d1/data/conf/type/__apt_source/manifest: {'__apt_source/nginx.org'}
/tmp/tmpv_r5stjc/68878079ef13f7cb21067fd9e16ae2d1/data/conf/type/__apt_source/manifest: {'__apt_source/elastic-5.x'}
(...)
```
Relevant lines from type manifests:
```
__logstash_install/manifest:require="__apt_source/elastic-5.x __package/openjdk-8-jdk-headless" __package logstash --state "$state"
__nginx_caching_proxy/manifest:require="__apt_source/nginx.org" __package nginx --state "$state"
```

darko-poljak:
@tom-ee This is just a warning. This is due to a dynamic nature of cdist dependency resolver. Some description/explanation is here: https://nico.schottelius.org/software/cdist/man/latest/man1/cdist.html#caveats

If there is no error then you can ignore this warning.
Is there an error?

tom-ee:
No error, _just_ the warning. Though this is not a parallel (neither `-j` nor `-p`) run.
Also the _result_ is correct (i.e. the index is properly updated).

My point is, that this warning is
a) not actionable, there is nothing I can do about it. The cdist user has to "know" that this is the normal behaviour. Why issue the warning at all in this (particular) case?
b) avoidable w/ the proposed changes.

I agree that the proposed change might result in additional runs of `apt-get update`, but as sources are not added on a regular basis this might be the lesser evil that issuing a (non actionable) warning in _every_ run.

darko-poljak:
@tom-ee
a) I think it is better that users know their current config run is of such case.
b) Proposed changes could impact performance, index update is expensive.

Since original author of `__apt_update_index` is @asteven and he and @telmich are BDFLs of cdist I cannot just change this `__apt_update_index` behavior.

@asteven , your thoughts?

darko-poljak:
Steven also pointed to break of backward compatibility.
@tom-ee  I am closing this PR. Perhaps new variant of __apt_source type would be more convenient.

PR #623, opened 2018-02-08
Remove update index type dep, call update index after adding new source.

darko-poljak:
__apt_source type is updated so that it does not use __apt_update_index, but, if necessary, it calls apt-get update if new source is added.

@tom-ee  Can you test it for your case?


tom-ee:
Works fine, thank you for the fast re-implementation!

darko-poljak:
@tom-ee Can you open new issue and continue messaging discussion there? This PR is going to be merged and closed. Thanks.

tom-ee:
@darko-poljak I created #624 for the messaging discussion.

PR #626, opened 2018-02-14
Consul refactoring

darko-poljak:
consul type is enhancened to download directly on servers instead of using staged_file.
Latest version 1.0.6 is added to type and set as default.

@telmich @asteven Any remarks?


telmich:
Misunderstanding from my side, @darko-poljak : @asteven needs __staged_file support ; what I meant is to add an option to download directly instead of using __staged_file instead of replacing it.

darko-poljak:
@telmich Refactored by leaving staged file support and adding new option for download and deploy directly on target machine.

telmich:
Looks good to me! Can be merged w/ changelog updated

PR #629, opened 2018-02-20
add `--state`-parameter and messaging to __package_dpkg

tom-ee:
- `--state` allows to remove a deb-package, if `--purge-if-absent` is
  specified the package is purged instead of "only" removed
- messaging was added
- man-page updated accordingly


cdist/conf/type/__package_dpkg/man.rst
```diff
@@ -12,30 +12,84 @@ This type is used on Debian and variants (like Ubuntu) to
...
+
+BOOLEAN PARAMETERS
+-------------------
+purge-if-absent
+    use `--purge` instead of just `--remove` for state=absent.
+
+
+BOOLEAN PARAMETERS
+------------------
+purge-if-absent
```
darko-poljak: @tom-ee It seems you have doubled BOOLEAN PARAMETERS section.

cdist/conf/type/__package_dpkg/man.rst
```diff
@@ -12,30 +12,84 @@ This type is used on Debian and variants (like Ubuntu) to
...
+    __package_dpkg baz_1.4_amd64.deb \
+        --source $__type/files/baz_1.4_amd64.deb \
+        --purge-if-absent \
+        --state "absent"
 
 
 SEE ALSO
 --------
 :strong:`cdist-type__package`\ (7)
+:strong:`dpkg-query`\ (1)
```
darko-poljak: @tom-ee List should be comma separated.

darko-poljak:
@tom-ee I assume you have tested all cases?

tom-ee:
Now I did -- and discovered the bug fixed in the already pushed commit.

Issue #632, opened 2018-02-22
[BETA] bug: error info lost with parallel (option -j)


AnotherKamila:
When running with `-b -j$whatever` and cdist fails, the stderr from types (NOT from the initial manifest) is not shown, which makes debugging types very hard.

# What I expect
if a type fails, its stderr should be displayed, just like when running without parallelization

# What happens
stderr not shown

# Steps to reproduce

**`type/__test_type/manifest`:**
```sh
echo "This is my stderr" >&2
exit 47
```

**`manifest/init`:**
```sh
__test_type whatever
```

Output without parallelization / `$ cdist config localhost`:
```
ERROR: localhost: Command failed: '/bin/sh -e /tmp/tmp8q6105_m/421aa90e079fa326b6494f812ad13e79/data/conf/type/__test_type/manifest'
return code: 47
---- BEGIN stdout ----
---- END stdout ----

Error processing object '__test_type/whatever'
===============================================
name: __test_type/whatever
path: /tmp/tmp8q6105_m/421aa90e079fa326b6494f812ad13e79/data/object/__test_type/whatever/.cdist-_u9hg_tz
source: /tmp/tmp8q6105_m/421aa90e079fa326b6494f812ad13e79/data/conf/manifest/init
type: /tmp/tmp8q6105_m/421aa90e079fa326b6494f812ad13e79/data/conf/type/__test_type

---- BEGIN manifest:stderr ----
This is my stderr

---- END manifest:stderr ----

ERROR: cdist: Failed to configure the following hosts: localhost
```

Output with parallelization / `$ cdist config -b -j4 localhost`:
```
ERROR: [28695]: localhost: Command failed: '/bin/sh -e /tmp/tmpvbovdxyv/421aa90e079fa326b6494f812ad13e79/data/conf/type/__test_type/manifest'
return code: 47
---- BEGIN stdout ----
---- END stdout ----
ERROR: cdist: Failed to configure the following hosts: localhost
```

Note that no error information is shown in this case.

tom-ee:
Might help with narrowing this down: Does the stderr make it (to stdout *sic*) if the output-stream-capture is disabled with `-S`?

AnotherKamila:
When running `cdist config -b -j4 -S localhost`, the stderr indeed is on stdout:
```
This is my stderr
[...]

PR #640, opened 2018-02-28
[BETA] bug: error info lost with parallel (option -j) #632

darko-poljak:


darko-poljak:
@AnotherKamila Merging this. If you find something else regarding this please open new issue.

PR #641, opened 2018-03-01
Add pre-exists state to __file

matthijskooijman:
Here's two (unrelated) changes I needed and which seem reusable. Commit messages should be self-explaining. If you prefer separate pull requests, let me know.

### DIFF
--- a/cdist/conf/type/__file/gencode-local
+++ b/cdist/conf/type/__file/gencode-local
@@ -25,6 +25,20 @@ type="$(cat "$__object/explorer/type")"
 
 [ "$state_should" = "exists" -a "$type" = "file" ] && exit 0 # nothing to do
 
+if [ "$state_should" = "pre-exists" ]; then
+   if [ -f "$__object/parameter/source" ]; then
+      echo "--source cannot be used with --state pre-exists"
+      exit 1
+   fi
+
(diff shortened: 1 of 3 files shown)


darko-poljak:
@matthijskooijman  Can you separate this PR into two? Thanks!

matthijskooijman:
Done!

telmich:
Interesting and no objections to the PR, but curious: when do you use pre-exists?

darko-poljak:
@matthijskooijman You could also describe pre-exists use case in man.rst :)

matthijskooijman:
My usecase is that I modify the bootup sequence of a beagleboard to start a specific script on startup, but that script should already be present on the default install of the beagleboard. To prevent referencing a script that does not exist and silently failing to boot, I want to check its existence, but if it does not exist, I don't know what the script should contain (and an empty file won't do either), so I can't use the normal exists.

I couldn't think of a nice example for the manpage that uses standard files, but I couldn't really think of anything that was generic enough. Suggestions are welcome :-)

darko-poljak:
So this is a way to check that file already exists at remote.

matthijskooijman:
Yup

telmich:
good to go

Issue #636, opened 2018-02-26
Inventory location inconsistency


matthijskooijman:
I've been looking at the inventory feature, and was a bit surprised about a few things:
 - The inventory location defaults to `~/.cdist/inventory`, whereas everything else seems to default to the in-distribution location (e.g `cdist/conf/manifest/init`). Is there any compelling for this reason?
 - An in-distibution location is used when `$HOME` is unset, but not when it is empty. This prevents switching to the other behaviour using `HOME= ./bin/cdist ...`, which isn't terribly convenient.
 - The in-distribution location used (when `$HOME` is unset) seems inconsistent: it is `cdist/inventory`, where I would expect `cdist/conf/inventory`, along with the rest of the config.

darko-poljak:
@matthijskooijman Inventory location defaults to ~/.cdist/inventory because there is nothing related to  inventory that comes with distribution, except implementation support. Also, in-distribution location cdist/inventory is used instead of cdist/conf/inventory because inventory is not tight coupled with manifests and types that are used for configuring hosts, and this is what is held in conf directory.
The HOME env var empty case will be fixed.

darko-poljak:
@matthijskooijman Can you test the following commit, related to empty HOME env var? https://github.com/ungleich/cdist/commit/2b5f1a01480a3e5ff4af0dc307387299a64db3a4

darko-poljak:
@matthijskooijman You can also use configuration file: https://www.nico.schottelius.org/software/cdist/man/latest/cdist-configuration.html
There is an option for inventory_dir. If you want a config file in git repo you can also have your home dir config file be link to the one in git repo.
Does this fit your needs?

matthijskooijman:
Yeah, that's also an option. I still need to pass a -g option explicitly. Would it make sense to automatically load a config file from the distribution, if it exists? e.g. `cdist/conf/cdist.conf` or something?

darko-poljak:
@matthijskooijman You can have user's configuration file (first one found of ~/.cdist.cfg, $XDG_CONFIG_HOME/cdist/cdist.cfg, in specified order) or
system-wide configuration file (/etc/cdist.cfg).

matthijskooijman:
Yeah, but none of these are inside my repository, so I can't make the repository self-contained (and it also won't work if I want to work with multiple self-contained repositories). Perhaps I'm thinking about this in the wrong way, but a in-distribution (default) config that is parsed would make sense, right?

[...]

PR #642, opened 2018-03-01
Support in-distribution config file

darko-poljak:
@matthijskooijman Can you test this?
It is related to last discussion for https://github.com/ungleich/cdist/issues/636 .


matthijskooijman:
This would do the trick for me, code looks sane too :-)

PR #643, opened 2018-03-02
Add __apt_default_release type

matthijskooijman:


cdist/conf/type/__apt_default_release/man.rst
```diff
@@ -0,0 +1,44 @@
...
+DESCRIPTION
+-----------
+Configure the default release for apt, using the APT::Default-Release
+configuration value. The name of the object sets the default release
+and can contain release name, codename or release version. Examples:
+'stable', 'testing', 'unstable', 'stretch', 'buster', '4.0', '5.0*'.
+
+REQUIRED PARAMETERS
+-------------------
+None.
```
darko-poljak: In description states that object name is release value and here is no parameter mentioned, but in the code you use release parameter and is in the list of required parameters. Can you fix this inconsistency?

cdist/conf/type/__apt_default_release/manifest
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh -e
+#
+# 2014 Steven Armstrong (steven-cdist at armstrong.cc)
+# 2017 Matthijs Kooijman (matthijs at stdin.nl)
```
darko-poljak: Did Steven participate in this type? If no then above lines are copy-paste relict :)

matthijskooijman: Well, a lot of the code is a copy-paste from `__apt_source` IIRC, so copyright on those lines is not mine to claim. It's not really big parts, so removing his copyright notice would probably be fine too, if that's what you prefer.

darko-poljak: @matthijskooijman Ok, you can leave this as is.

cdist/conf/type/__apt_default_release/man.rst
matthijskooijman: Ah, I originally wanted to use the object name, but didn't know a singleton has no name. I probably forgot to update the manpage afterwards.

In any case, I fixed this now. I added a fixup commit instead of amending the previous commit, for easier review, but they should be squashed before merging (but I think you usually manually merge anyway, right?).

cdist/conf/type/__apt_default_release/man.rst
```diff
@@ -0,0 +1,46 @@
...
+
+
+EXAMPLES
+--------
+
+.. code-block:: sh
+
+    __apt_default_release stretch
```
darko-poljak: One more request. In this example required release parameter is missing :)
After that I think this PR is ready for merging.

matthijskooijman: Heh, good thing you're reviewing this :-P

Fixed now.

PR #647, opened 2018-03-11
New __prometheus_server, __prometheus_alertmanager, __grafana_dashboard

AnotherKamila:
Updated the types to work with packages instead of `go get`, removed `__daemontools` dependency and cleaned up. Tested on Devuan Ascii (and Devuan Jessie, but see below).

Note: Due to significant differences between Prometheus 1.x and 2.x, this type assumes 2.x and fails if your packages install 1.x. Prometheus 2.x is available in ascii backports (use `--install-from-backports`), but not in jessie (even backports have 1.x). Therefore only Devuan Ascii or newer is supported unless you add a package source with newer Prometheus yourself.

---
This change is [Reviewable](https://reviewable.io/reviews/ungleich/cdist/647)


AnotherKamila:
Note that the last commit depends on #648.

cdist/conf/type/__prometheus_alertmanager/manifest
```diff
@@ -1,34 +1,63 @@
...
+
+# TODO this is a bug in the init script, patching it like this is awful and it should be reported
+require="$require $require_pkg" \
+__key_value alertmanager_fix_init_script --file /etc/init.d/prometheus-alertmanager \
+	--key "NAME" --value "prometheus-alertmanager" --delimiter "=" \
+	--onchange "service prometheus-alertmanager restart"
+
+##### CONFIGURE #############################################################
+
+FLAGS="--storage.path $storage_path --data.retention $(($retention_days*24))h --web.listen-address [::]:9093"
```
telmich: I like IPv6, but does [::] also listen on IPv4?

AnotherKamila: Yes, the kernel listens on all when told `[::]`.

darko-poljak:
@AnotherKamila Can you remove __key_value changes in this PR?

AnotherKamila:
@darko-poljak Why? The commit was cherry-picked, so it should merge just fine...

darko-poljak:
@AnotherKamila It's kind of unclean :)

AnotherKamila:
Done

darko-poljak:
@AnotherKamila Thanks! ;)

PR #650, opened 2018-03-13
Add new type __docker_stack

lubo:


cdist/conf/type/__docker_stack/gencode-remote
```diff
@@ -0,0 +1,63 @@
...
+		if [ "${compose_file}" = "-" ]; then
+			compose_file="${__object}/stdin"
+		fi
+
+		cat <<-EOF
+		compose_file="\$(mktemp cdist.XXXXXXXXXX)"
+
+		base64 -d > "\${compose_file}" << eof
+		$(base64 "${compose_file}")
+		eof
```
darko-poljak: @lubo I think that eof should be on the beginning of line.

cdist/conf/type/__docker_stack/gencode-remote
```diff
@@ -0,0 +1,63 @@
...
+
+		docker stack deploy --compose-file "\${compose_file}" \
+			--prune ${stack}
+
+		rm "\${compose_file}"
+		EOF
```
darko-poljak: Here too, EOF should be on the beginning of the line.

cdist/conf/type/__docker_stack/gencode-remote
lubo: Content of this heredoc is being printed to stdout here, so the `eof` will be on the beginning of line.

cdist/conf/type/__docker_stack/gencode-remote
lubo: Why? Shellcheck doesn't report any problems on this line.

darko-poljak:
Line with it has leading spaces so it is not considered as here-doc end
tag.

Darko

cdist/conf/type/__docker_stack/gencode-remote
lubo: I have used `<<-` instead of `<<` and tabs instead of spaces, though. Could you point me to a reference that says this syntax is incorrect, please?

darko-poljak:
@lubo Sorry, my fault. Yes, because of `<<-` leading tabs are stripped from line containing delimiter too.

PR #658, opened 2018-04-30
Add new type __docker_secret

lubo:


cdist/conf/type/__docker_secret/explorer/secret-exists
```diff
@@ -0,0 +1,25 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with cdist. If not, see <http://www.gnu.org/licenses/>.
+#
+
+if docker secret ls | grep -q " ${__object_id:?} "; then
+	echo 1
+else
+	echo 0
```
telmich: I suggest to use "yes"/"no" here, as it does not require the user to jump from gencode-remote to the explorer to review what 0 or 1 means

lubo: Changed.

PR #661, opened 2018-05-07
Rewrite __letsencrypt_cert type

lubo:
This commit **adds** these features:

 * Ability to expand existing certificate
 * Ability to manage object state
 * Ability to obtain test certificate
 * Ability to promote test certificate to production
 * Ability to specify custom certificate name
 * Ability to specify multiple domains per certificate
 * Ability to use Certbot in standalone mode
 * Messaging

This commit **removes** the following features:

 * Install Certbot
 * Install certificate renewal cron job

These tasks are not a concern of this type and should be left to the
user or other types in order to maintain flexibility.

**Note:** `renew-hook` optional parameter was removed. Therefore,
these changes are breaking.


lubo:
@AnotherKamila @darko-poljak @telmich Your input would be highly appreciated.

telmich:
How will you manage servers if you remove the renew hook? i.e. nginx needs to be restarted, when the certificate is renewed

lubo:
`renew-hook` parameter was removed because this type no longer creates the cron job which was using it.

The user should add the cron job on his/her own, outside of this type. If the code stayed in this type, we'd have to expose many more options to make it as flexible as necessary, while we can achieve the same or better flexibility when the code is not present in this type at all.

AnotherKamila:
Personally, I disagree with the removal of the features: I believe that
- certbot should be installed from the repositories by default and a boolean parameter (such as `--no-certbot-install`) could be provided for cases where something non-default needs to happen
- cron job should be installed by default, and either a boolean parameter to turn this off or an optional `--cron-raw-command` (or similar) parameter should be made available for special cases

Both of these make it much easier for the user to do the default thing, which I would guesstimate is at least 90% of cases. This way the simplicity of the common case would be preserved, which is awesome, and the special cases would also be doable, which is also awesome. The cost of doing this is a slightly more complex type, but I believe that it would be a good tradeoff.

lubo:
All right. I've brought the two back with slight changes:

 * If `certbot` binary can't be found on the target machine, this type will try to install it using preferred, system-specific method (i.e., using package manager or pip). Otherwise, nothing happens. I've added an explorer that check if the binary exists (not a system-specific package), so no additional parameters nor user intervention should be required in any case.
* Installation of the cron job can be enabled using `--automatic-renewal` parameter. It is disabled by default, because it can be “broken” by default. Nowadays, many systems do not have cron installed (nor it can be installed, in several cases, such as CoreOS) or using cron is not the preferred way of scheduling jobs. I think it makes more sense to let the user enable it when he/she needs it, rather them having him/her disable it when he/she finds out it's broken.

I'd say this is a fair compromise. Any thoughts?

AnotherKamila:
Sounds good to me. I'm still a bit grumpy about breaking backwards compatibility, but I can live with it if it's trivial to fix (which it now is thanks to the `--automatic-renewal` flag).

What do the actual maintainers think? :D @darko-poljak @telmich

darko-poljak:
Now it's better.
@telmich Any additional remarks/suggestions?

lubo:
@telmich Ping.

telmich:
Well reasoned, convinced to break my configurations.

Darko, we need at least a minor version change and a hint in the changes file for this.

Merge, please.

PR #666, opened 2018-05-27
Fix __letsencrypt_cert type: use object id as domain

darko-poljak:
and optional additional domains through --domain param.

@telmich @lubo Please comment.


lubo:
Currently, the object ID is used as certificate name, which is an arbitrary string that is part of the certificate filename in the file system. The user currently is (and should be) free to choose any other certificate name than a valid FQDN. However, with your changes, if he/she didn't choose the certificate name to be a valid FQDN, like `example.com-0001`, Certbot'd most certainly crash with an error, complaining about the supplied domain name not being a valid FQDN.

Also, I failed to understand what exactly is the motion behind these changes. Is it to partially restore backward compatibility? Could you elaborate on that, please?

telmich:
@lubo We have a lot of types that by default use the object_id. If you only want one certificate, you don't need --domain. It's a cdist idiom that by default you require least effort from the user.

If you have multiple domains then you have to use something that allows multiplication, like --domain a --domain b.

And yes, the original behaviour of the type was to take object_id as the domain name, because that makes sense. I am not sure, if we should even stick with --domain or name it --alternate-name, to map it more closely to what it actually is.

darko-poljak:
@telmich @lubo How about a compromise?
If a user does not specify --domain argument then object id is used as a domain.
If --domain is specified then object id is not used as a domain name so cert name can be set to arbitrary string.
There are types that work in such a way, but currently, I cannot recall an example.

lubo:
@darko-poljak I can agree to that, sounds good.

@telmich I see. Also, I did copy the naming from Certbot without thinking too much about it, but I don't agree that `--domain` parameter should be renamed. Although Let's Encrypt limits itself to providing certificates for valid FQDNs only, SANs aren't limited to domain names. So, I think the naming is actually spot-on.

lubo:
@darko-poljak Btw, I'd also really like to review the code when you change it, if it's OK with you.

cdist/conf/type/__letsencrypt_cert/gencode-remote
```diff
@@ -16,6 +16,12 @@ case "${state}" in
 	;;
 	present)
 		requested_domains="${__object}/parameter/domain"
+        tmp_domains=$(mktemp domain.cdist.XXXXXXXXXX)
+        if [ -f "${requested_domains}" ]; then
+            cp "${requested_domains}" "${tmp_domains}"
+        else
+            echo "$__object_id" >> "${tmp_domains}"
+        fi
```
lubo: Why don't you just write `${__object_id}` to `${requested_domains}` if the file does not exist? I believe using a temporary file is overkill. Also, the script uses tabs instead of spaces for indentation.

cdist/conf/type/__letsencrypt_cert/gencode-remote
```diff
@@ -26,10 +32,10 @@ case "${state}" in
 			existing_domains="${__object}/explorer/certificate-domains"
 			certificate_is_test=$(cat "${__object}/explorer/certificate-is-test")
 
-			sort -uo "${requested_domains}" "${requested_domains}"
+			sort -uo "${tmp_domains}" "${tmp_domains}" -
```
lubo: The trailing hyphen is unnecessary.

darko-poljak:
@telmich @lubo I have updated implementation.
object id is used as domain only if domain parameter is not specified.
This way user can specify arbitrary string for cert name.

cdist/conf/type/__letsencrypt_cert/gencode-remote
darko-poljak: Ah yes, that one I forgot to remove :)

cdist/conf/type/__letsencrypt_cert/gencode-remote
darko-poljak: But we want to have original parameter values in .cdist/cache, don't we?

darko-poljak: @telmich What would you prefer here, to temp or not to temp? :)

lubo: @darko-poljak Not sure if I made myself clear. So, I was wrong at first and I think that you're right, a temporary file should be used. What I don't like is naming of the variable that holds path to the temporary file. I think that `${requested_domains}` is more suitable than `${tmp_domains}` for the temporary file and the current `${requested_domains}` should be renamed to something else, or perhaps just get removed (it's used only twice).

darko-poljak: @lubo Aaa, I misunderstood you, sorry.

darko-poljak: @lubo Renamed.

darko-poljak:
@telmich Any remarks?

cdist/conf/type/__letsencrypt_cert/man.rst
```diff
@@ -44,6 +42,9 @@ OPTIONAL MULTIPLE PARAMETERS
 renew-hook
     Renew hook command directly passed to Certbot in cron job.
 
+domain
+    Domains to be included in the certificate.
```
telmich: I suggest to add the hint regarding skipping object_id here, too.

cdist/conf/type/__letsencrypt_cert/man.rst
```diff
@@ -68,6 +69,14 @@ remove
...
+
+    __letsencrypt_cert example.com \
+        --admin-email root@example.com \
+        --automatic-renewal \
+        --renew-hook "service nginx reload" \
+        --webroot /data/letsencrypt/root
+
 .. code-block:: sh
 
     __letsencrypt_cert example.com \
```
telmich: Add a comment here that example.com needs to be included again at --domain

darko-poljak:
@telmich man page updated

PR #669, opened 2018-06-07
add __acl: Basic wrapper around setfacl

4nd3r:
@telmich asked:
> Why do you think recursive should be the default?

to keep things simple and because mostly, when dealing with ACL entries, you want recursive operation.

code will get very complex and error prone, if we want to differentiate between recursive and non-recursive ACL entries because `getfacl` doesn't tell (without `-R` and even with it, parsing it sanely is rather difficult, YMMV) if ACL entry is also on subdirs and files. so as first iteration I tried to get type which would cover 90% of my use cases. will see about rest of 10% in the future :-).

also about `--remove` flag - maybe some other phrase would be better, but the idea is that all destructive operations should be explicit.

### DIFF
--- a/cdist/conf/type/__acl/explorer/acl_is
+++ b/cdist/conf/type/__acl/explorer/acl_is
@@ -1,0 +1,23 @@
+#!/bin/sh -e
+#
+# 2018 Ander Punnar (ander-at-kvlt-dot-ee)
+#
+# This file is part of cdist.
+#
+# cdist is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# cdist is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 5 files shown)


4nd3r:
also, i found some bugs. please do not merge yet.

4nd3r:
ok, i think it's now somewhat better. please review and comment.

4nd3r:
wow, getting some fresh air helps.

so, i should add `--recursive` and `--default`. this way i can resolve another 5% of my use cases.

for example set non-default group to parent directroy, but treat known subdirectories differently.

will test that tomorrow.

darko-poljak:
@4nd3r I suggest to add safety check for supported OSes. For example, FreeBSD setfacl does not support -R option.

4nd3r:
> I suggest to add safety check for supported OSes.

and solaris doesn't support `-x`. should check should be whitelist or blacklist?

i think it's easier just take all linux dists from __explorer/os and write them down.

4nd3r:
i think it's good now. any more comments?

darko-poljak:
@4nd3r Also check OS in explorer for getfacl. E.g. FreeBSD getfacl has neither -c nor -p option.

4nd3r:
-c and -p are not needed actually, because grep :>

darko-poljak:
@4nd3r One more thing. If commit history isn't necessary, please rebase :)

@telmich Any remarks on this new type?

4nd3r:
rebased

PR #672, opened 2018-06-15
__sysctl: support sysctl.d

### LINKED ISSUE #671: __sysctl: /etc/sysctl.conf is not being read by systemd
__sysctl writes the configuration to `/etc/sysctl.conf`, which is not being read by systemd anymore. According to ArchWiki:

>From version 207 and 21x, systemd only applies settings from `/etc/sysctl.d/*.conf` and `/usr/lib/sysctl.d/*.conf`. If you had customized `/etc/sysctl.conf`, you need to rename it as `/etc/sysctl.d/99-sysctl.conf`. If you had e.g. `/etc/sysctl.d/foo`, you need to rename it to `/etc/sysctl.d/foo.conf`.

We should probably default to `/etc/sysctl.d/99-sysctl.conf` on systems where systemd is used.

darko-poljak:
Resolves #671.

@asteven What do you think of this simple implementation? It adheres you suggestions from #671.

@lubo Any remarks?

### DIFF
--- a/cdist/conf/type/__sysctl/explorer/conf-path
+++ b/cdist/conf/type/__sysctl/explorer/conf-path
@@ -1,6 +1,6 @@
-#!/bin/sh -e
+#!/bin/sh
 #
-# 2014 Steven Armstrong (steven-cdist at armstrong.cc)
+# 2018 Darko Poljak (darko.poljak at gmail.com)
 #
 # This file is part of cdist.
 #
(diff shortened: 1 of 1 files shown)


cdist/conf/type/__sysctl/explorer/conf-path
```diff
@@ -0,0 +1,21 @@
...
+# cdist is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with cdist. If not, see <http://www.gnu.org/licenses/>.
+#
+
+[ -d "/etc/sysctl.d" ] && echo "/etc/sysctl.d/99-Z-sysctl-cdist.conf" || echo "/etc/sysctl.conf"
```
telmich: I just wanted to complain about @asteven 's claim that we always want to be last in this directory, but I can't, he's totally right. I think *if* an existing /etc/sysctl.d means that files are being read in there, then this patch is a beauty.

lubo: if/else would be safer and it'd also improve readability.

asteven:
LGTM

Issue #675, opened 2018-06-20
__line crashes in dry-run mode when file does not exist yet


thriqon:
Suppose the following snippet in a manifest:

```
__file /tmp/test --state exists
require="__file/tmp/test" __line blah --file /tmp/test --line "blah is there"
```

and `/tmp/test` does not exist on the target host.

This works fine when invoking `cdist config` normally, but crashes in the explorer phase when running it with `--dry-run`.

----

This is because the [explorer](https://github.com/ungleich/cdist/blob/master/cdist/conf/type/__line/explorer/state) invokes `awk` directly without testing file presence first and handling that gracefully. `__line` still crashes (probably) if the file does not exist when running normally, but that is a different topic (and a design decision).

PR #676, opened 2018-06-20
Handle missing files with __line (dry-run)

thriqon:
Second commit is optional.

### DIFF
--- a/cdist/conf/type/__line/explorer/state
+++ b/cdist/conf/type/__line/explorer/state
@@ -39,6 +39,11 @@ else
    file="/$__object_id"
 fi
 
+if [ ! -f "$file" ]; then
+  echo "file_missing"
+  exit 0
+fi
+
 awk -v position="$position" -v needle="$needle" '
 function _find(_text, _pattern) {
    if (needle == "regex") {


asteven:
Thanks for the patch.
But to me it looks like you are using it wrongly.

If the file does not exist we should fail. Question here is if we can fail in a more user friendly way then just letting awk die hard.

We can't just create the file if it is not there. We have no clue how user/group/permissions should be.

The proper way to use this is by using the __file type to ensure the file exists.

e.g.
```
__file /tmp/my-file \
   --state present
require="__file/tmp/my-file" \
   __line my-line \
      --file /tmp/my-file \
      --line 'hello world'
```

Like this the explorer works fine, because it is run after the __file type has done it's job.
Also no need to create the file in gencode-remote.

asteven:
Sorry, just saw that I missed the context from #675.

thriqon:
Yes, it usually works correctly, but not with --dry-run, as the file is not created then.

I'll drop the second commit, fine with me.

On 20 June 2018 12:35:37 GMT+02:00, Steven Armstrong <notifications@github.com> wrote:
>Thanks for the patch.
>But to me it looks like you are using it wrongly.
>
>If the file does not exist we should fail. Question here is if we can
>fail in a more user friendly way then just letting awk die hard.
>
>We can't just create the file if it is not there. We have no clue how
>user/group/permissions should be.
>
>The proper way to use this is by using the __file type to ensure the
>file exists.
>
>e.g.
>```
>__file /tmp/my-file \
>   --state present
>require="__file/tmp/my-file" \
>   __line my-line \
>      --file /tmp/my-file \
>      --line 'hello world'
>```
>
>Like this the explorer works fine, because it is run after the __file
>type has done it's job.
>Also no need to create the file in gencode-remote.

-- 
mail@jonasw.de
https://jonasw.de/

thriqon:
I've dropped the second commit. Could you take a look again, please?

PR #678, opened 2018-07-11
Improve __systemd_unit

adam-dej:
This PR adds support for masking systemd unit files, and fixes minor issues with type `__systemd_unit`.


darko-poljak:
@lubo Since you introduced this type, and I am neither user nor have deeper knowledge about systemd, can you review this changes? Thanks!

PR #686, opened 2018-09-01
Check if certbot exists before using it. Fixes #685.

### LINKED ISSUE #685: __letsencrypt_cert certificate-* explorers fail when certbot does not exist
When certbot is not yet installed on target host, __letsencrypt_cert explorers certificate-* result in `certbot: not found`.

In certificate-* explorers we should use certbot-path explorer to check if it exists and do logic accordingly.

darko-poljak:
This PR fixes #685.

@lubo @asteven @telmich  Any remarks?


darko-poljak:
@lubo Changed to tabs.

PR #689, opened 2018-09-22
don't duplicate Remote::mkdir in explorer transfer

myeisha:
Remote::transfer contains a call to mkdir(destination) if the source is
a directory. since the destination must also be a directory we can omit
extra mkdir calls in Explorer.

this saves about 10% on my manifests in sequential mode, and about 6% in
parallel mode.


darko-poljak:
@myeisha Indeed. Thanks!

PR #690, opened 2018-09-22
transfer all files of a directory at once

myeisha:
instead of calling the copy command once per file in a directory (eg a
type explorer dir), call the copy command only once with all files of
the directory.

batch copy saves 20% of dry-run time on my test manifest in sequential
mode, and 15% in parallel mode.

### DIFF
--- a/cdist/exec/remote.py
+++ b/cdist/exec/remote.py
@@ -203,46 +203,20 @@ class Remote(object):
                     os.remove(tarpath)
                     used_archiving = True
             if not used_archiving:
-                if jobs:
-                    self._transfer_dir_parallel(source, destination, jobs)
-                else:
-                    self._transfer_dir_sequential(source, destination)
+                self._transfer_dir(source, destination)
         elif jobs:
             raise cdist.Error("Source {} is not a directory".format(source))
         else:
             self._transfer_file(source, destination)
 
-    def _transfer_dir_commands(self, source, destination):
+    def _transfer_dir(self, source, destination):
+        command = self._copy.split()
         for f in glob.glob1(source, '*'):
-            command = self._copy.split()
             path = os.path.join(source, f)
-            command.extend([path, '{0}:{1}'.format(
-                _wrap_addr(self.target_host[0]), destination)])
-            yield command
-
-    def _transfer_dir_sequential(self, source, destination):
-        for command in self._transfer_dir_commands(source, destination):
-            self._run_command(command)
-
-    def _transfer_dir_parallel(self, source, destination, jobs):
-        """Transfer a directory to the remote side in parallel mode."""
-        self.log.debug("Remote transfer in {} parallel jobs".format(
-            jobs))
-        self.log.trace("Multiprocessing start method is {}".format(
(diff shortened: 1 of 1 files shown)


darko-poljak:
@telmich @asteven Any remarks on this one?

asteven:
Not sure about this one. How would this play with non-default remote-copy scripts?
Can this even work with the default remote_copy?

```
[21:50:51] crius:~% mkdir /tmp/test-dir
[21:50:54] crius:~% touch /tmp/test-dir/{first,second,third}
[21:51:03] crius:~% la /tmp/test-dir 
total 0
drwxr-xr-x  2 sar  sar   100 Sep 22 21:51 .
drwxrwxrwt 49 root root 1080 Sep 22 21:50 ..
-rw-r--r--  1 sar  sar     0 Sep 22 21:51 first
-rw-r--r--  1 sar  sar     0 Sep 22 21:51 second
-rw-r--r--  1 sar  sar     0 Sep 22 21:51 third
[21:51:05] crius:~% 
[21:51:06] crius:~% scp -o User=root /tmp/test-dir localhost:/tmp/foobar
/tmp/test-dir: not a regular file
[21:51:32] crius:~% 
```

Or am I missing something here?

myeisha:
it works perfectly fine for me with the default settings. the change expands all filenames and stuffs them into a single invocatoin rather than using one invocation per file, that should work for any command that "behaves like scp". recursive copy was something i did think about, but that would change semantics (and needs more pathname munging).

darko-poljak:
Just to clarify to all, and to myself, this is essentially this form:
```
scp file1 file2 file3 file4 host:destination
```
instead of
```
scp file host:destination
```
for each file.

@telmich @asteven  As we are not using `scp` `-r` option because some other custom remote copy could not support it, do we expect the above to be supported?
@asteven Can you test this with your more complex custom remote copy commands (+chroot and similar) before merging this into master?

asteven:
> 
> @telmich @asteven As we are not using `scp` `-r` option because some other custom remote copy could not support it, do we expect the above to be supported?

I don't have a problem to support this. It's 'ssh/scp like' and also works with plain 'cp'.

> @asteven Can you test this with your more complex custom remote copy commands (+chroot and similar) before merging this into master?

My remote-copy script will work with this. All it does is it turns this:

```
scp file1 file2 file3 file4 host:destination
```
into this:
```
scp file1 file2 file3 file4 host:/path/to/chroot/destination
```

darko-poljak:
>> @telmich @asteven As we are not using scp -r option because some other custom remote copy could not support it, do we expect the above to be supported?

> I don't have a problem to support this. It's 'ssh/scp like' and also works with plain 'cp'.

`cp` `-r` option is not by POSIX :D , `-r` option is historical practice on BSD and BSD-derived systems.

Linux/GNU cp supports it the same as `-R`.

So `cp -r` may not work as expected on all systems.

asteven:
I ment I don't have a problem to support this pull request. I don't have any need nor interest for ```scp -r```.

Issue #540, opened 2017-07-03
make everything pass shellcheck


4nd3r:
even if we exclude SC2154 (something is referenced but not assigned), then others should be dealt with in some point. yes, that's lot of work :>

thoughts on this?

```
$ find . -type f | while read -r l; do file "$l" | grep -Fq 'POSIX shell script' && shellcheck -f gcc "$l"; done | sed -nr 's/.+:[0-9]+:\s(.+)$/\1/p' | sort | uniq -c | sort -nr
    221 note: Double quote to prevent globbing and word splitting. [SC2086]
    196 warning: __object is referenced but not assigned. [SC2154]
    172 warning: __object_id is referenced but not assigned. [SC2154]
     41 warning: __global is referenced but not assigned. [SC2154]
     40 warning: __type is referenced but not assigned. [SC2154]
     30 warning: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined. [SC2166]
     17 warning: __explorer is referenced but not assigned. [SC2154]
     17 error: Iterating over ls output is fragile. Use globs. [SC2045]
     14 warning: __object_name is referenced but not assigned. [SC2154]
     13 warning: Prefer [ p ] || [ q ] as [ p -o q ] is not well defined. [SC2166]
     12 warning: Quote this to prevent word splitting. [SC2046]
     11 warning: __messages_out is referenced but not assigned. [SC2154]
     11 warning: Declare and assign separately to avoid masking return values. [SC2155]
     11 note: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead. [SC2002]
     10 warning: __target_host is referenced but not assigned. [SC2154]
      9 note: Useless echo? Instead of 'echo $(cmd)', just use 'cmd'. [SC2005]
      9 note: $/${} is unnecessary on arithmetic variables. [SC2004]
      7 note: Expressions don't expand in single quotes, use double quotes for that. [SC2016]
      6 warning: In POSIX sh, 'local' is undefined. [SC2039]
      5 warning: Remove surrounding $() to avoid executing output. [SC2091]
      5 warning: __remote_exec is referenced but not assigned. [SC2154]
      5 warning: Quotes/backslashes in this variable will not be respected. [SC2090]
      5 warning: In POSIX sh, echo flags are undefined. [SC2039]
      5 note: Use find instead of ls to better handle non-alphanumeric filenames. [SC2012]
      4 warning: __remote_copy is referenced but not assigned. [SC2154]
[...]

PR #693, opened 2018-10-01
Fix SC2166 (and/or operators in test) on all manfest/gencode/explorers

thriqon:
Relevant documentation: https://github.com/koalaman/shellcheck/wiki/SC2166

Apparently, the famous `-a` and `-o` are not well-defined by POSIX, and warned against by shellcheck.

shellcheck stats: 499 -> 452
relates to: #540


darko-poljak:
@telmich @asteven If you have no remarks then I will merge this.

asteven:
LGTM

PR #696, opened 2018-10-02
[WIP] Fix SC2086 on all scripts

thriqon:
Relevant documentation: https://github.com/koalaman/shellcheck/wiki/SC2086

relates to: #540

This commit addresses 241 separate cases of missing quotes around
variables.

----
This PR conflicts with #694. I'm aware of that and will fix the code as soon as that one is merged. Before that, this PR should not be merged.


darko-poljak:
@thriqon Sorry, I meant on the style in your and other PR #697. There are changes like
```
echo "add-apt-repository '$name'"
```

And yours are like 
```
echo "add-apt-repository \"$name\""
```

thriqon:
In theory, this could make a difference, as the expansion of `$name` is re-expanded on the target host in this PR, whereas in #697 it is not.

I'd say this PR is less breaking in this regard, but is this really behaviour that is a) expected and b) desired? In retrospect, I'd choose the single-quote style as well.

asteven:
Personally I don't use echo much anymore these days.
With printf most of the quoting issues simply do not exist.

e.g.
```
printf 'add-apt-repository %s\n' "$name"
```

asteven:
But if I have to choose one of the above I'd go for the single-quote style as well.

darko-poljak:
@thriqon Can you review your changes, compare it with #697 and change escaped quotes with single- and double- quotes combinations where appropriate?

thriqon:
@darko-poljak, I did that. Could you please review?

cdist/conf/type/__directory/gencode-remote
```diff
@@ -57,18 +57,18 @@ get_current_value() {
 }
 
 set_group() {
-    echo chgrp $recursive \"$1\" \"$destination\"
-    echo chgrp $recursive $1 >> "$__messages_out"
+    echo "chgrp '$recursive' '$1' '$destination'"
```
tom-ee: This will fail if `$recursive` is empty.

cdist/conf/type/__package_yum/gencode-remote
```diff
@@ -60,10 +60,10 @@ fi
 
 case "$state_should" in
     present)
-        echo yum $opts install \"$install_name\"
+        echo "yum $opts install \"$install_name\""
```
tom-ee: We could switch the inner `\"` to `'` here too.

cdist/conf/type/__rvm/explorer/state
```diff
@@ -28,7 +28,7 @@ if [ "$user" = "root" ]; then
         echo absent
     fi
 else
-    if su - $user -c "[ -d \"\$HOME/.rvm\" ]" ; then
+    if su - "$user" -c "[ -d '\$HOME/.rvm' ]" ; then
```
tom-ee: This breaks the intended behaviour -- '$HOME' will not expand in the su-environment.

cdist/conf/type/__directory/gencode-remote
thriqon: Unquoted `$recursive` (safe, since it only ever contains nothing or `-R`, and is only set by us).

cdist/conf/type/__package_yum/gencode-remote
thriqon: fixed, was due to be fixed yesterday evening, but somehow...

cdist/conf/type/__rvm/explorer/state
thriqon: reverted.

thriqon:
Ready for review (again).

tom-ee:
LGTM

darko-poljak:
LGTM too.
@asteven Can you too get some time to review these changes?
If you don't spot something fishy then I will merge this PR.

PR #699, opened 2018-10-03
Fix SC2039

darko-poljak:
SC2039

In POSIX sh, something is undefined.

https://github.com/koalaman/shellcheck/wiki/SC2039

__package_pkg_freebsd I left untouched because this type is meant to be run on FreeBSD anyway and in its shell those are defined.
Also, in type files scripts, where shebang is bash I left non-POSIX _something_ unchanged.

### DIFF
--- a/cdist/conf/type/__docker/manifest
+++ b/cdist/conf/type/__docker/manifest
@@ -24,7 +24,7 @@ state=$(cat "$__object/parameter/state")
 
 case "$os" in
     centos)
-        if (source "$__global/explorer/os_release" && [ "${VERSION_ID}" = "7" ]); then
+        if (. "$__global/explorer/os_release" && [ "${VERSION_ID}" = "7" ]); then
           __yum_repo docker-ce-stable \
             --name 'Docker CE Stable' \
             --baseurl "https://download.docker.com/linux/centos/7/\$basearch/stable" \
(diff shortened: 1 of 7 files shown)


cdist/conf/type/__docker/manifest
```diff
@@ -24,7 +24,7 @@ state=$(cat "$__object/parameter/state")
 
 case "$os" in
     centos)
-        if (source "$__global/explorer/os_release" && [ "${VERSION_ID}" = "7" ]); then
+        if (. "$__global/explorer/os_release" && [ "${VERSION_ID}" = "7" ]); then
```
asteven: $__global/explorer/os_release is supposed to executed, not sourced.

darko-poljak: @asteven Aren't **source** and **.**(dot) the same and only dot is defined by POSIX?

asteven: They are. But explorers are **executables**. There is no need to source them at all.

darko-poljak: @asteven  As I understand in the above code "$__global/explorer/os_release" is the output of os_release explorer after it has been run on target host. This output is read and executed using *dot* in the current env and then it is checked if VERSION_ID equals 7. I understand that the output of os_release contains such shell compatible variable definition.

darko-poljak: @asteven Ping

PR #700, opened 2018-10-03
Remove unused variables (SC2034)

thriqon:
When doing this, I found two locations where the wrong (?) variables were used (`ip` instead of `cur_ip`). I left the code in; please take a look a there. I found two unrelated instances, where manifest code was setting parameter defaults. I haven't seen this before, is this still needed?

Solves SC2034

references #540.

- [x] check `type/__install_umount/manifest`
- [x] check `type/__jail_freebsd9/gencode-remote`
- [x] check `type/__jail_freebsd10/gencode-remote`
- [X] check `type/__install_reboot/manifest`


darko-poljak:
@jdguffey Hello! Can you take a look at jail_freebsd types and resolve the above?

asteven:
The 'setting parameter defaults' in manifest is a relict from a time long before there where defaults for parameters. This can be removed/changed as long as the functionality remains the same.

thriqon:
I removed those manifests, and replaced it with an explicit default in `__install_umount`. `__install_remount` didn't use this parameter anyways (at all!), so I just dropped it entirely.

darko-poljak:
@thriqon I suggest to fix __jail_freebsd* files so that cur_ip is used in for loop instead of ip.
I would say that was an intention in the first place.
After that I think this PR can be merged.
@jdguffey If this is not the case please fix this in the separate PR.

PR #701, opened 2018-10-04
fix SC2148 (missing shebang) and make executable

tom-ee:
see <https://github.com/koalaman/shellcheck/wiki/SC2148>: `Tips depend on target shell and yours is unknown. Add a shebang.`

This adds missing shebang to all scripts and add's the missing executable flag.

RFC: I added `set -e` to these [explorer shebangs]. Is there a reason why only manifests and gencode-scripts have `set -e` and explorers do not?

### DIFF
--- a/cdist/conf/explorer/disks
+++ b/cdist/conf/explorer/disks
@@ -1,2 +1,3 @@
+#!/bin/sh -e
 cd /dev
 echo sd? hd? vd?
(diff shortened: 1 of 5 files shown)


asteven:
(Global) Explorers should always be able to fail gracefully. Not sure if they have a chance to do that if we `set -e` them.

tom-ee:
I wasn't aware of that -- `set -e` removed on the explorers.

tom-ee:
Most files that needed a new CWD are either very short or the `cd` was inside a sub-shell.
A 3rd pair of eyes is welcome of course.

darko-poljak:
@tom-ee I will merge this one. If any additional work is needed that will be done in separate PR.
Let's go step by step.

PR #702, opened 2018-10-04
fix another SC2166 (test `[ .. -a.. ]` replaced by `[ .. ] && ]`)

tom-ee:
one missing SC2166, <https://github.com/koalaman/shellcheck/wiki/SC2166>


darko-poljak:
@tom-ee Thanks!

PR #703, opened 2018-10-04
fix SC2045 (use globs instead of `ls`)

tom-ee:
`Iterating over ls output is fragile. Use globs.` <https://github.com/koalaman/shellcheck/wiki/SC2045>

This also unifies the 2 styles that were in place ...


darko-poljak:
@asteven I will review/check for side effects after changing CWD too.

darko-poljak:
I didn't notice any side effect, I am merging this.

PR #707, opened 2018-10-04
Disable SC2016 check.

darko-poljak:
SC2016

Expressions don't expand in single quotes, use double quotes for that.

https://github.com/koalaman/shellcheck/wiki/SC2016

But here this is intended.


darko-poljak:
@tom-ee @thriqon @asteven Do you agree?

PR #708, opened 2018-10-05
Fix for various problems (shellcheck)

darko-poljak:


darko-poljak:
@asteven @tom-ee @thriqon Please review.

cdist/conf/type/__jail_freebsd10/gencode-remote
```diff
@@ -111,6 +111,7 @@ startJail() {
    fi
 }
 
+# shellcheck disable=SC2120
```
thriqon: This error is legitimate (and is caused by a bug).
In line 125 and line 130 the code mentions `$3` which AFAICS is expanded by the heredoc, not by awk.

cdist/conf/type/__jail_freebsd10/gencode-remote
```diff
@@ -347,6 +348,7 @@ if [ "$present" = "EXISTS" ]; then   # The jail currently exists
       exit 0
    else   # The jail is not supposed to exist
       stopJail
+      # shellcheck disable=SC2119
```
thriqon: I expect this to vanish, if the above bug is fixed.

cdist/conf/type/__jail_freebsd9/gencode-remote
```diff
@@ -116,6 +116,7 @@ startJail() {
    fi
 }
 
+# shellcheck disable=SC2120
```
thriqon: Equivalent problem to `freebsd10`, I suppose (didn't check, though).

cdist/conf/type/__jail_freebsd9/gencode-remote
```diff
@@ -342,6 +343,7 @@ if [ "$present" = "EXISTS" ]; then   # The jail currently exists
       exit 0
    else   # The jail is not supposed to exist
       stopJail
+      # shellcheck disable=SC2119
```
thriqon: see above.

cdist/conf/type/__consul_template_template/manifest
```diff
@@ -52,6 +53,7 @@ for param in *; do
                --owner root --group root --mode 640 \
                --source "$source" \
                --state "$state"
+         # shellcheck disable=SC2030
          export require="__file${destination}"
```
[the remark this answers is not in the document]
asteven: It's honored by exporting it into the environment.
The next __type invocation will have and honor it

**export** require="__file${destination}"

darko-poljak:
@asteven @tom-ee @thriqon Ready for review again.

cdist/conf/type/__install_partition_msdos_apply/gencode-remote
```diff
@@ -21,11 +21,11 @@
 #set -x
 
 die() {
-   echo "[__install_partition_msdos_apply] $@" >&2
+   echo "[__install_partition_msdos_apply] $*" >&2
```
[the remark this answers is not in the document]
darko-poljak: I followed https://github.com/koalaman/shellcheck/wiki/SC2145.

darko-poljak:
@thriqon Do you have any other remark/suggestion?

cdist/conf/type/__consul_template_template/manifest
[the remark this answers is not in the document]
thriqon: Yes, I know @asteven, but there is no more __type invocation in this subshell. And the require won't propagate out of that subshell (as shellcheck is indeed correct to mention).

thriqon:
@darko-poljak, nothing new, just the (still open) `export require` question.

cdist/conf/type/__consul_template_template/manifest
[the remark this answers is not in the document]
darko-poljak: @thriqon As I see, inside subshell there is a for loop, for each param, and inside loop body __file is called.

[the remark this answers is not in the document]
darko-poljak: cdist records this requirement when next __file type is called.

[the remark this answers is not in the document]
thriqon: As far as I'm able to understand the underlying intent, this exported require is expected to be available in line 70. Isn't that correct? It won't be available there! This is what shellcheck tells us.

Anyways, the next `__file` invocation in another loop iteration does *not* use this require as well, as it has an explicit `require="..."` in front of it (it won't hit that branch again anyways, as there won't be two distinct files with the same name.

cdist/conf/type/__consul_template_template/manifest
```diff
@@ -66,7 +69,7 @@ for param in *; do
    esac
 done
 printf '}\n'
-) | \
+} | \
```
darko-poljak: @thriqon @asteven Using { } instead of ( ) does not work here either, because of pipe it is still executed in subshell. So I moved variable assignment before this. Please review.

darko-poljak: @asteven Can you verify that this works as expected?

asteven: I will fix it when/if it breaks something. Move along.

PR #710, opened 2018-10-06
Shellcheck makefile target

darko-poljak:
@telmich @asteven @tom-ee @thriqon I added shellcheck makefile target and shellcheck step for release target.

I am ignoring SC2154 for variables starting with __ since such variables are cdist environment variables.


darko-poljak:
Great! Thanks

Darko

PR #711, opened 2018-10-06
Shellcheck rest

darko-poljak:
# SC1091
Not following: (error message here)
https://github.com/koalaman/shellcheck/wiki/SC1091

# SC1117
Backslash is literal in "\n". Prefer explicit escaping: "\\\\n".
https://github.com/koalaman/shellcheck/wiki/SC1117

# SC2154
var is referenced but not assigned.
https://github.com/koalaman/shellcheck/wiki/SC2154
But ignoring __\<name\> vars since such vars are cdist env vars.

@thriqon @tom-ee @asteven Please review.

When resolving SC2154 I found bugs and resolved them, but one I marked as **TODO** since
I am not sure how it should be resolved.


cdist/conf/type/__process/gencode-remote
```diff
@@ -52,6 +52,7 @@ case "$state_should" in
         if [ -f "$__object/parameter/stop" ]; then
             cat "$__object/parameter/stop"
         else
+            # TODO: what should be runs variable?
             echo kill "${runs}"
```
darko-poljak: @telmich @asteven Can you help here? I am not sure what **runs** should be.

darko-poljak: And I didn't study the type thoroughly.

PR #712, opened 2018-10-07
Fix problems found by sellcheck in type explorers.

darko-poljak:
@asteven @thriqon @tom-ee Here are another fixes, for type explorers.
This is still work in progress, see code comments below.

### DIFF
--- a/cdist/conf/type/__link/explorer/state
+++ b/cdist/conf/type/__link/explorer/state
@@ -33,6 +33,7 @@ destination_dir="${destination%/*}"
 case "$type" in
    symbolic)
       cd "$destination_dir" || exit 1
+      # TODO SC2012: use readlink or something?
       source_is=$(ls -l "$destination" | sed 's/.*-> //g')
       if [ -h "$destination" ]; then
          # ignore trailing slashes for comparison
@@ -52,7 +53,9 @@ case "$type" in
          echo sourcemissing
          exit 0
       fi
+      # TODO SC2012: use stat?
       destination_inode=$(ls -i "$destination" | awk '{print $1}')
+      # TODO SC2012: use stat?
       source_inode=$(ls -i "$source" | awk '{print $1}')
       if [ "$destination_inode" -eq "$source_inode" ]; then
          echo present
--- a/cdist/conf/type/__link/explorer/state
+++ b/cdist/conf/type/__link/explorer/state
@@ -33,8 +33,7 @@ destination_dir="${destination%/*}"
 case "$type" in
    symbolic)
       cd "$destination_dir" || exit 1
-      # TODO SC2012: use readlink or something?
-      source_is=$(ls -l "$destination" | sed 's/.*-> //g')
+      source_is=$(readlink "$destination")
       if [ -h "$destination" ]; then
          # ignore trailing slashes for comparison
          if [ "${source_is%/}" = "${source%/}" ]; then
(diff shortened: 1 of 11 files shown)


cdist/conf/type/__start_on_boot/explorer/state
```diff
@@ -38,12 +38,27 @@ if [ "$init" = 'systemd' ]; then
 else
     case "$os" in
         debian|openwrt|devuan)
-            state="present"
```
darko-poljak: The changes below are due to https://github.com/koalaman/shellcheck/wiki/SC2144

cdist/conf/type/__link/explorer/state
```diff
@@ -33,6 +33,7 @@ destination_dir="${destination%/*}"
 case "$type" in
    symbolic)
       cd "$destination_dir" || exit 1
```
darko-poljak: https://github.com/koalaman/shellcheck/wiki/SC2012

Should we use something different here instead of ls?

The same for below TODOs.

cdist/conf/type/__link/explorer/state
```diff
@@ -33,6 +33,7 @@ destination_dir="${destination%/*}"
...
+      # TODO SC2012: use readlink or something?
       source_is=$(ls -l "$destination" | sed 's/.*-> //g')
       if [ -h "$destination" ]; then
```
darko-poljak: https://github.com/koalaman/shellcheck/wiki/SC2012

Should we use something different here instead of ls?

The same for TODOs in code below this line.

cdist/conf/type/__start_on_boot/explorer/state
```diff
@@ -38,12 +38,27 @@ if [ "$init" = 'systemd' ]; then
...
+            for file in "/etc/rc$runlevel.d/S"??"$name"
+            do
+                if [ -f "$file" ]
+                then
+                    state="present"
+                    break
+                fi
+            done
+
             [ -f "/etc/init/${name}.conf" ] && state="present"
```
darko-poljak: These changes are due to https://github.com/koalaman/shellcheck/wiki/SC2144

cdist/conf/type/__link/explorer/state
```diff
@@ -52,7 +53,9 @@ case "$type" in
          echo sourcemissing
          exit 0
       fi
+      # TODO SC2012: use stat?
       destination_inode=$(ls -i "$destination" | awk '{print $1}')
```
tom-ee: `stat --format '%i' "$destination` is perfect and saves one fork (`stat -f '$i'` on FreeBSD)

cdist/conf/type/__link/explorer/type
```diff
@@ -31,6 +31,7 @@ elif [ -f "$destination" ]; then
    type="$(cat "$__object/parameter/type")"
    case "$type" in
       hard)
+         # TODO SC2012: use stat?
          link_count=$(ls -l "$destination" | awk '{ print $2 }')
```
tom-ee: `stat --format '%h' "$destination"` is perfect on Linux -- FreeBSD according to <https://www.freebsd.org/cgi/man.cgi?query=stat&apropos=0&sektion=0&manpath=FreeBSD+11.2-RELEASE+and+Ports&arch=default&format=html> `stat -f '%l' "$destination`

cdist/conf/type/__link/explorer/state
tom-ee: Thinking about this it's probably not worth the effort to change this code as `/usr/bin/stat` does not seem to be POSIX. Adapting for all supported OSes make the code harder to read and maintain so we are probably better with disabling SC2012 in this case (instead of using `stat`).
This does _not_ affect the `readlink`-replacement.

darko-poljak: @tom-ee I was thinking the same.

darko-poljak:
@asteven @thriqon @tom-ee I resolved TODOs. Some are changed in favor of readlink and others are ignored, as @tom-ee commented.
So no more _work in progress_, it is ready for final reviewing.

PR #715, opened 2018-10-11
Improve  user type

tom-ee:
3 improvements for __user:

* redirect stdout and error of `userdel` to get rid of annoying output
* handle typos in `--state`-parameter gracefully (previously `--state presetn` or similar typos would have delete the user
* message user removal (w/ and w/o remove-home)

manpage updated accordingly.

### DIFF
--- a/cdist/conf/type/__user/gencode-remote
+++ b/cdist/conf/type/__user/gencode-remote
@@ -135,10 +135,10 @@ elif [ "$state" = "absent" ]; then
     if grep -q "^${name}:" "$__object/explorer/passwd"; then
         #user exists, but state != present, so delete it
         if [ -f "$__object/parameter/remove-home" ]; then
-            printf "userdel -r %s >/dev/null 2>&1\n" "${name}"
+            printf "userdel -r '%s' >/dev/null 2>&1\n" "${name}"
 	    echo "userdel -r" >> "$__messages_out"
         else
-            printf "userdel %s >/dev/null 2>&1\n" "${name}"
+            printf "userdel '%s' >/dev/null 2>&1\n" "${name}"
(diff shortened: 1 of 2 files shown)


cdist/conf/type/__user/gencode-remote
```diff
@@ -130,13 +131,17 @@ if [ "$state" = "present" ]; then
           echo useradd "$@" "$name"
        fi
     fi
-else
+elif [ "$state" = "absent" ]; then
     if grep -q "^${name}:" "$__object/explorer/passwd"; then
         #user exists, but state != present, so delete it
         if [ -f "$__object/parameter/remove-home" ]; then
-            echo userdel -r "${name}"
+            printf "userdel -r %s >/dev/null 2>&1\n" "${name}"
```
thriqon: I'd suggest single quotes here, so the name arrives quoted on the target.

cdist/conf/type/__user/gencode-remote
```diff
@@ -130,13 +131,17 @@ if [ "$state" = "present" ]; then
...
+	    echo "userdel -r" >> "$__messages_out"
         else
-            echo userdel "${name}"
+            printf "userdel %s >/dev/null 2>&1\n" "${name}"
```
thriqon: see above, single quotes?

cdist/conf/type/__user/gencode-remote
tom-ee: I was hesitant about this one, as users can (should?) not contain whitespace -- but better safe that sorry here ...

darko-poljak:
@asteven @telmich Any remarks/suggestions?

Issue #716, opened 2018-10-12
Object conflict where there should not be one - or wrong error message


telmich:
~~~~
[8:29] line:~% cdist config -vvc ~/vcs/ungleich-dot-cdist/ -p10 red1.place5.ungleich.ch         
VERBOSE: cdist: version 4.10.3-22-g383af673                                
WARNING: [11468]: red1.place5.ungleich.ch: Could not derive host_name for red1.place5.ungleich.ch, $host_name will be empty. Error is: [Errno 1] Unknown host
INFO: [11468]: red1.place5.ungleich.ch: Starting configuration run
ERROR: [11468]: red1.place5.ungleich.ch: Linking manifest /home/nico/vcs/ungleich-dot-cdist/manifest/.#dcl to /tmp/tmpe0cvk44v/9613ea11c03c21bc404d3d251acd217e/data/conf/manifest/.#dcl failed: [Errno 17] File exists: '/home/nico/vcs/ungleich-dot-cdist/manifest/.#dcl' -> '/tmp/tmpe0cvk44v/9613ea11c03c21bc404d3d251acd217e/data/conf/manifest/.#dcl'
VERBOSE: [11468]: config: Total processing time for 1 host(s): 0.2670779228210449
[8:29] line:~% cdist config -vvc ~/vcs/ungleich-dot-cdist/ -p10 red1.place5.ungleich.ch
VERBOSE: cdist: version 4.10.3-22-g383af673
WARNING: [11565]: red1.place5.ungleich.ch: Could not derive host_name for red1.place5.ungleich.ch, $host_name will be empty. Error is: [Errno 1] Unknown host
INFO: [11565]: red1.place5.ungleich.ch: Starting configuration run
VERBOSE: [11565]: red1.place5.ungleich.ch: Running global explorers
VERBOSE: [11565]: red1.place5.ungleich.ch: Running initial manifest /tmp/tmptirtsbvg/9613ea11c03c21bc404d3d251acd217e/data/conf/manifest/init
VERBOSE: [11565]: red1.place5.ungleich.ch: Preparing object __ungleich_ntp/
VERBOSE: [11565]: red1.place5.ungleich.ch: Running manifest and explorers for __ungleich_ntp/
VERBOSE: [11565]: red1.place5.ungleich.ch: Running type explorers for <CdistType __ungleich_ntp>
[...]

PR #718, opened 2018-10-12
Fix DirectoryDict getitem.

darko-poljak:
This fixes https://github.com/ungleich/cdist/issues/716.
The problem is in DirectoryDict getitem, see https://github.com/ungleich/cdist/issues/716 for details.


asteven:
Really not sure about this one. Honestly I do not remember how exactly this works.
Assume tests are still passing with your changes?

darko-poljak:
@asteven I wouldn't create a PR if `make test` fails :)

The only place in code where `cdist_object.parameters` are referenced and at the same place it is
relevant what is `cdist_object.parameters` concrete value is when checking for object conflicting parameters. Except this, parameter files are only handled in shell code from types.

In setitem it is detected if value is a sequence, and then each value is written on its line in file.
This works for multiple parameters.
Single value is written as single line.
For boolean parameters there is no value, and this is handled with
`# ensure file ends with a single newline` part, so an empty line in file.
```
    def __setitem__(self, key, value):
        try:
            with open(os.path.join(self.path, key), "w") as fd:
                if (not hasattr(value, 'strip') and
                    (hasattr(value, '__getitem__') or
                     hasattr(value, '__iter__'))):
                    # if it looks like a sequence and quacks like a sequence,
                    # it is a sequence
                    for v in value:
                        fd.write(str(v) + '\n')
                else:
                    fd.write(str(value))
                    # ensure file ends with a single newline
                    if value and value[-1] != '\n':
                        fd.write('\n')
```
But in getitem only the whole file content was read.

My change handles those three cases in getitem.
If read value is empty then '' is expected (boolean parameters).
If there is only one line in file then this is the case when some parameter has single value.
If there are multiple lines in file then this is the case for multiple value parameter.
```
                if not value:
                    return ''
                elif len(value) == 1:
                    return value[0]
                else:
                    return value
```

asteven:
> @asteven I wouldn't create a PR if `make test` fails :)
> 

Off course. Sorry.

> [..]
> My change handles those three cases in getitem.
> If read value is empty then '' is expected (boolean parameters).
> If there is only one line in file then this is the case when some parameter has single value.
> If there are multiple lines in file then this is the case for multiple value parameter.
> 

The only case I'm not sure is the last one. What about:
- there could be multiple parameters but currently there is only one

But as you said. If all this is only used to compare emulator params vs existing object params there should not be a problem.

darko-poljak:
> > @asteven I wouldn't create a PR if `make test` fails :)
> 
> Off course. Sorry.

I could have forgotten though :)

> 
> > [..]
> > My change handles those three cases in getitem.
> > If read value is empty then '' is expected (boolean parameters).
> > If there is only one line in file then this is the case when some parameter has single value.
> > If there are multiple lines in file then this is the case for multiple value parameter.
> 
> The only case I'm not sure is the last one. What about:
> 
>     * there could be multiple parameters but currently there is only one
Hm... I will check.
> 
> 
> But as you said. If all this is only used to compare emulator params vs existing object params there should not be a problem.

darko-poljak:
@asteven You were right.

```
$ cat ~/.cdist/manifest/test
__ungleich_monit ungleich --config ungleich
__ungleich_monit ungleich --config ungleich
$ ./bin/cdist config -vv -i ~/.cdist/manifest/test -n 185.203.112.26
VERBOSE: cdist: version 4.10.3-127-g151edc63
INFO: 185.203.112.26: Starting dry run
VERBOSE: 185.203.112.26: Running global explorers
VERBOSE: 185.203.112.26: Running initial manifest /home/darko/.cdist/manifest/test
ERROR: 185.203.112.26: /bin/sh -e /home/darko/.cdist/manifest/test: ['/bin/sh', '-e', '/home/darko/.cdist/manifest/test']

Error processing initial manifest
=================================
path: /home/darko/.cdist/manifest/test

init:stderr
-----------
ERROR: cdist: Object __ungleich_monit/ungleich already exists with conflicting parameters:
/home/darko/.cdist/manifest/test: {'state': 'present', 'config': 'ungleich'}
/home/darko/.cdist/manifest/test: {'state': 'present', 'config': ['ungleich']}


VERBOSE: config: Total processing time for 1 host(s): 4.922122955322266
ERROR: cdist: Failed to configure the following hosts: 185.203.112.26
```
Do you have any suggestion?
Should we, for object parameters comparison, consider `[ 'value' ]` the same as `'value'`?

asteven:
> 
> Do you have any suggestion?

The problem is that deep down in fsproperty you don't have any context. You don't know what kind of data your dealing with.

> Should we, for object parameters comparison, consider `[ 'value' ]` the same as `'value'`?

Not nice, but since that's the only place we seem to use it would be an option. I'd try to not invest to much time here. Make it work, somehow, move on.

darko-poljak:
@asteven Please take a quick look. Last commit added params adaptation based on cdist_type context before checking for conflicting parameters.

PR #723, opened 2018-11-26
type: Add new type __ping

oliv3r:
When starting with cdist, the documentation and examples can be
overwheliming. Further more, sometimes one would only want to do a
simple test to see if the basics are working. One such example currently
is to create a simple remote file. While this is a nice and simple
example, an even simpler example would be to just see if we can 'ping'
the host.

Other configuration management tools also seem to have this as a basic
starting command.

This thus allows to do:
echo __ping | cdist config --initial-manifest - <host>

as the most basic, most simple command to 'test' things with, without
having lingering files to boot.

Signed-off-by: Olliver Schinagl <oliver@schinagl.nl>

### DIFF
--- a/cdist/conf/type/__ping/gencode-remote
+++ b/cdist/conf/type/__ping/gencode-remote
@@ -1,0 +1,12 @@
+#!/bin/sh -e
+#
+# Copyright (C) 2018 Olliver Schinagl <oliver@schinagl.nl>
+#
+# SPDX-License-Identifier: GPL-3.0+
+#
+
+set -eu
+
+echo "echo 'pong'"
+
+exit 0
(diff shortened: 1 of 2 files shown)


4nd3r:
we do have `__cdistmarker` for simple testing, but why not `__ping` too.

but there are some strange temporary looking `*.un~` files in commit...

oliv3r:
Whoops, you are right. I accidentally added my vim backup files. I opened #725 to address this in the future :)
Also, __cdistmarker is slightly different, it actually changes the target and does a little more ...

darko-poljak:
@telmich @asteven Your thoughts?

telmich:
Sounds good to me, I think @oliv3r 's use case is valid and the type is nicely small. :+1:

PR #724, opened 2018-11-26
Add editorconfig

oliv3r:
With editorconfig, users have their editor setup automatically for the
codestyle of this project. While some editors have support for editor
config built in by default, others require a plugin. In both cases, it
is fully optional.

Signed-off-by: Olliver Schinagl <oliver@schinagl.nl>

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+indent_style = space
+insert_final_newline = true


darko-poljak:
@oliv3r Not sure about that, it's up to the user to setup his/her editors. I think this is not part of the cdist.

4nd3r:
i'm in between.

cdist doesn't enforce strict coding style, yet. maybe some day? then this kind of thing would be useful.

oliv3r:
Sorry for the late reply,

just note that this does not enforce anything. It just informs the editor about a few trivial things (such as setting the editor properly up to do line endings on the last line, or that when i press tab in my editor, it inserts 4 spaces.

It does not enforce anything, and if you don't have an editor that supports this feature, it doesn't do anything. 

I'm okay if not wanting to have this, but to be fair, you then should also strip the UTF-8 marker as can be seen here https://github.com/ungleich/cdist/blob/6c8014b407cf28ee54764ffc6182cb28a580dc66/bin/cdist#L2 as you are basically doing the same but doing it on a per file basis ...

just some food for thought.

PR #730, opened 2018-12-14
All package types send messages upon installation/removal

sideeffect42:
I extended all package types to post an "installed" message when a package has been installed (or updated) and a "removed" message when a package has been removed.

Also, I completely reworked the `__package_pkg_openbsd` type.


telmich:
Big fan of the messaging aspect! Did not review openbsd type yet, but love your approach!

darko-poljak:
@sideeffect42 Could you split this PR into new two of them and close this one? One new for adding messages and one new for `__package_pkg_openbsd` rework only.
Thanks!

PR #732, opened 2018-12-14
All package types send messages upon installation/removal

sideeffect42:
I extended all package types to post an "installed" message when a package has been installed (or upgraded) and a "removed" message when a package has been removed.


darko-poljak:
@sideeffect42 I cherry-pick-ed that commit.

PR #734, opened 2019-01-22
readd the use of echo in my explorers

dheule:
this is nessessary for
 remove blanks in output (strip)
 catch exit codes of zypper (for example if no repo is defined at all)

==> without this echo's the type's doesn't work at all.

### DIFF
--- a/cdist/conf/type/__zypper_repo/explorer/all_repo_ids
+++ b/cdist/conf/type/__zypper_repo/explorer/all_repo_ids
@@ -21,4 +21,4 @@
 # Retrieve all repo id nummbers - parsed zypper output
 #
 #
-zypper lr | cut -d'|' -f 1 | grep -E '^[0-9]'
+echo $(zypper lr | cut -d'|' -f 1 | grep -E '^[0-9]')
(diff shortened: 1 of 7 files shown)


thriqon:
Note that this change violates shellcheck rules: [SC2005](https://github.com/koalaman/shellcheck/wiki/SC2005)

asteven:
@thriqon please note that @dheule is the primary author of the __zypper_* types which were broken by our shell-check'ifying. We should try to fix this asap in a way that you both can live with.

@thriqon how would we mark those uses of echo so that shell-check ignores them?

asteven:
Never mind, found it:

```
# shellcheck disable=SC2005
echo ...
```

@dheule: Could you please mark the echo's like that so it doesn't break our shell-check run?

thriqon:
@asteven, exactly what I was typing right now ;-)

darko-poljak:
@dheule After you make this change I plan to release new bugfix version.

darko-poljak:
@dheule ping

dheule:
sorry, overseen the mails, i will implement the # shellcheck disable=SC2005
please stand by for update on this pr

dheule:
pr is now updated to prefix needed echos with # shellcheck disable=SC2005

darko-poljak:
@dheule Thanks.

PR #648, opened 2018-03-11
__key_value: --onchange parameter added

AnotherKamila:
(cherry picked from commit a7881414ef5cef7cd5e92b5710f0c3bc23505c6c)

---
This change is [Reviewable](https://reviewable.io/reviews/ungleich/cdist/648)

PR #743, opened 2019-03-28
add --onchange to __file

4nd3r:
because having to use `__config_file` just for that feels redundant

### DIFF
--- a/cdist/conf/type/__file/gencode-remote
+++ b/cdist/conf/type/__file/gencode-remote
@@ -23,7 +23,7 @@ destination="/$__object_id"
 state_should="$(cat "$__object/parameter/state")"
 type="$(cat "$__object/explorer/type")"
 stat_file="$__object/explorer/stat"
-
+fire_onchange=''
 
 get_current_value() {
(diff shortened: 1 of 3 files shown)


cdist/conf/type/__file/gencode-remote
```diff
@@ -91,3 +91,9 @@ case "$state_should" in
         exit 1
     ;;
 esac
+
+if [ -f "$__object/parameter/onchange" ]; then
+   if [ -s "$__messages_out" ]; then
```
darko-poljak: @4nd3r How do you know that `__file` type wrote to messages_out?
It could do nothing, but types before it wrote to messages_out.
I think you should grep for something or indicate that onchange should happen.

4nd3r: right, you are correct. stole idea from #648.

darko-poljak:
@4nd3r One last thing. Can you rebase this branch/PR so that it contains only one commit, for cleaner history?

4nd3r:
sure and thanks

PR #744, opened 2019-03-29
new type: __clean_path

4nd3r:
too often i need to make sure that some directories are empty or directory only contains files i know it should contain.

one good example is apache2 packaging and `conf-enabled`, `mods-enabled` and `sites-enabled` directories.

hopefully this somewhat dangerous type is useful to other people.

### DIFF
--- a/cdist/conf/type/__clean_path/explorer/list
+++ b/cdist/conf/type/__clean_path/explorer/list
@@ -1,0 +1,39 @@
+#!/bin/sh -e
+#
+# 2019 Ander Punnar (ander-at-kvlt-dot-ee)
+#
+# This file is part of cdist.
+#
+# cdist is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# cdist is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with cdist. If not, see <http://www.gnu.org/licenses/>.
+#
+
+path="/$__object_id"
+
+if [ ! -d "$path" ]
+then
+    echo "$path is not a directory" >&2
+    exit 1
+fi
+
+pattern="$( cat "$__object/parameter/pattern" )"
+
+if [ -f "$__object/parameter/exclude" ]
+then
+    exclude="$( cat "$__object/parameter/exclude" )"
+
+    find "$path" -mindepth 1 -maxdepth 1 -regex "$pattern" \
+        -and -not -regex "$exclude"
+else
+    find "$path" -mindepth 1 -maxdepth 1 -regex "$pattern"
+fi
--- a/cdist/conf/type/__clean_path/gencode-remote
+++ b/cdist/conf/type/__clean_path/gencode-remote
@@ -1,0 +1,48 @@
+#!/bin/sh -e
+#
+# 2019 Ander Punnar (ander-at-kvlt-dot-ee)
+#
+# This file is part of cdist.
+#
+# cdist is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# cdist is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with cdist. If not, see <http://www.gnu.org/licenses/>.
+#
+
+[ ! -s "$__object/explorer/list" ] && exit 0
+
+path="/$__object_id"
+
+pattern="$( cat "$__object/parameter/pattern" )"
+
+if [ -f "$__object/parameter/exclude" ]
+then
+    exclude="$( cat "$__object/parameter/exclude" )"
+
+    echo "find '$path' -mindepth 1 -maxdepth 1 -regex '$pattern'" \
+        "-and -not -regex '$exclude'" \
+        '-exec rm -rf {} \;'
+else
+    echo "find '$path' -mindepth 1 -maxdepth 1 -regex '$pattern'" \
+        '-exec rm -rf {} \;'
+fi
+
+while read -r f
+do
+    echo "removed '$f'" >> "$__messages_out"
+done \
+< "$__object/explorer/list"
+
+if [ -f "$__object/parameter/onchange" ]
+then
+    cat "$__object/parameter/onchange"
+fi
--- a/cdist/conf/type/__clean_path/man.rst
+++ b/cdist/conf/type/__clean_path/man.rst
@@ -1,0 +1,60 @@
+cdist-type__clean_path(7)
+=========================
+
+NAME
+----
+cdist-type__clean_path - Remove files and directories which match the pattern.
+
+
+DESCRIPTION
+-----------
+Remove files and directories which match the pattern.
+
+Provided path (as __object_id) must be a directory.
+
+Patterns are passed to ``find``'s ``-regex`` - see ``find(1)`` for more details.
+
+Look up of files and directories is non-recursive (``-maxdepth 1``).
+
+Parent directory is excluded (``-mindepth 1``).
+
+This type is not POSIX compatible (sorry, Solaris users).
+
+
+REQUIRED PARAMETERS
+-------------------
+pattern
+   Pattern of files which are removed from path.
+
+
+OPTIONAL PARAMETERS
+-------------------
+exclude
+   Pattern of files which are excluded from removal.
+
+onchange
+   The code to run if files or directories were removed.
+
+
+EXAMPLES
+--------
+
+.. code-block:: sh
+
+    __clean_path /etc/apache2/conf-enabled \
+        --pattern '.+' \
+        --exclude '.+\(charset\.conf\|security\.conf\)' \
+        --onchange 'service apache2 restart'
+
+
+AUTHORS
+-------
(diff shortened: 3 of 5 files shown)


tom-ee:
Why not just `__rsync` the desired directory (even an empty one)?On Mar 29, 2019 10:07, Ander Punnar <notifications@github.com> wrote:too often i need to make sure that some directories are empty or directory only contains files i know it should contain.
one good example is apache2 packaging and conf-enabled, mods-enabled and sites-enabled directories.
hopefully this somewhat dangerous type is useful to other people.



You can view, comment on, or merge this pull request online at:
  https://github.com/ungleich/cdist/pull/744

Commit Summary
add __rm

File Changes

    A
    cdist/conf/type/__rm/explorer/list
    (42)
  
    A
    cdist/conf/type/__rm/gencode-remote
    (28)
  
    A
    cdist/conf/type/__rm/man.rst
    (49)
  
    A
    cdist/conf/type/__rm/parameter/optional
    (2)
  
    A
    cdist/conf/type/__rm/parameter/required
    (1)
  

Patch Links:
https://github.com/ungleich/cdist/pull/744.patchhttps://github.com/ungleich/cdist/pull/744.diff

4nd3r:
> Why not just `__rsync` the desired directory (even an empty one)?

`__rsync` is "beta" and doesn't really suite my needs (no `--onchange` and always generates code even if not needed) and ideologies :wink:

cdist/conf/type/__rm/explorer/list
```diff
@@ -0,0 +1,42 @@
...
+    path="$( cat "$__object/parameter/path" )"
+else
+    path="/$__object_id"
+fi
+
+[ ! -d "$path" ] && exit 0
+
+remove="$( cat "$__object/parameter/remove" )"
+
+list="$( find "$path" -mindepth 1 -maxdepth 1 | grep -E "$remove" )"
```
darko-poljak: What if file contains spaces?

Why not refactor this type and use `find`'s `-path  pattern` or `-name pattern` options along with `-exec ...`?

4nd3r: hmm, yes, using `find` sounds saner. check next commit.

telmich:
Even though __rm is pretty self explanatory (good!) I somewhat object the name, as it describes an action versus a state. Problem is, I did not find a better name like __empty_directory or __empty_file ... maybe __empty?

Anyway, low prio the name suggestion, but if possible a self explanatory state would be preferred, if it doesn't exist, continue with __rm

4nd3r:
well, the type doesn't always empty the directory because of `--exclude`.

maybe `__remove` to be more explicit?

4nd3r:
or `__remove_files_and_or_directories`

4nd3r:
or just `__remove_files`, because in unix, directory is just special type of file :wink:

cdist/conf/type/__rm/explorer/list
```diff
@@ -0,0 +1,40 @@
...
+[ ! -d "$path" ] && exit 0
+
+remove="$( cat "$__object/parameter/remove" )"
+
+if [ -f "$__object/parameter/exclude" ]
+then
+    exclude="$( cat "$__object/parameter/exclude" )"
+
+    find "$path" -mindepth 1 -maxdepth 1 -path "$remove" \
+        -and -not -path "$exclude"
```
darko-poljak: @4nd3r mindepth and maxdepth are not by POSIX :)
@telmich Do we allow it anyway?

cdist/conf/type/__rm/man.rst
```diff
@@ -0,0 +1,52 @@
...
+-----------
+This type will DELETE your files and directories which match the pattern.
+
+Patterns are passed to ``find``'s ``-path`` - see find(1) for more details.
+
+
+REQUIRED PARAMETERS
+-------------------
+remove
+   Remove pattern.
```
darko-poljak: Can you better explain what "Remove pattern" is? Even rename the parameter?

cdist/conf/type/__rm/explorer/list
darko-poljak: `-and`, `-not` too

4nd3r: `-mindepth` is needed to exclude $PWD from list and `-maxdepth` because recursive actions are devil's tools. `-and` and `-not` are just convenience and we can probably get rid of them, but not sure about depth params.

cdist/conf/type/__rm/man.rst
4nd3r: rename to what? at first i used `--pattern`, but it didn't feel right :disappointed:

cdist/conf/type/__rm/explorer/list
4nd3r: and i actually forgot to add note about non-recursivness in manual... :sleepy:

4nd3r: even openbsd's `find` support `-mindepth`, `-maxdepth`, `-and` and `-not` so i think we are in the clear here, sans non-recursion note, if you guys agree that we don't want recursivness here...

cdist/conf/type/__rm/man.rst
darko-poljak: Is this parameter's value path pattern or file (inside path) pattern?
Depending on that it could be named as `remove-file-pattern`/`remove-path-pattern`.
I am not sure, but `remove` is confusing.

darko-poljak:
@4nd3r @telmich Re naming this type: what about something like `__clean_path` ?

cdist/conf/type/__rm/explorer/list
darko-poljak: What about `-prune`? ;)

darko-poljak: Something like
```
$ find .
.
./a
./a/b
./a/b/c
./a/b/c/d
./a/b/c/d/e
./a/b/c/d/e/f
./a/b/c/d/d1
./a/b/c/d/d2
./a/b/c/d/a1
./a/b/c/d/a2
./a/b/c/c1
./a/b/c/c2
./a/b/c/a1
./a/b/c/a2
./a/b/b1
./a/b/b2
./a/b/a1
./a/b/a2
./a/a1
./a/a2
$ find a/b/c -path a/b/c -o -type d -prune -o -type f -a -name "a*" -print
a/b/c/a1
a/b/c/a2
```

4nd3r:
+1 for `__clean_path` since I don't have better suggestion

cdist/conf/type/__rm/man.rst
4nd3r: then `--pattern`? should be clear enough and nicely short.

cdist/conf/type/__rm/explorer/list
4nd3r: with `-prune` it's not readable anymore, imho. unless you consume some hallucinogens.

darko-poljak: Ha ha ha ha ha ha 
@telmich Your thoughts on this one?

cdist/conf/type/__rm/man.rst
darko-poljak: Fine with me, better than remove.

cdist/conf/type/__clean_path/explorer/list
```diff
@@ -0,0 +1,35 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with cdist. If not, see <http://www.gnu.org/licenses/>.
+#
+
+path="/$__object_id"
+
+[ ! -d "$path" ] && exit 0
```
4nd3r: maybe `exit 1`?

cdist/conf/type/__rm/explorer/list
4nd3r: had one beer and i'm now trying to understand why `-prune`...

we use only one command and not two (second being `rm`)?

maybe we are splitting hairs here? :thinking:

4nd3r: okay, i think i understand now, a bit. since we are only operating on one level, then we should use `-name`.

4nd3r: replaced `-path` with `-name`

darko-poljak: `-prune` The primary shall always evaluate as true; it shall cause find not to descend the current pathname if it is a directory.

It can be used instead of `mindepth`, `maxdepth` which are not supported by POSIX.

But that is something @telmich  or @asteven should decide :)

4nd3r: new morning and new ideas: `-name` should be replaced with `-regex`.

and i'll take look at `-prune`...

darko-poljak:
@4nd3r This type should remove specified files in the directory, not descending into subdirectories, right?
Why not then `cd` into path and do the operation there?
You can also skip executing `find` both in the explorer and in `code`.
`find` could `-exec` some script that emits message and then removes the file.
Then explorer is even not needed.

What do you think?

darko-poljak:
Some POSIX examples can be found in 
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/find.html
like

```
The following command:

find . ! -name . -prune -name '*.old' -exec \
    sh -c 'mv "$@" ../old/' sh {} +

performs the same task as:

mv ./*.old ./.old ./.*.old ../old/
```

cdist/conf/type/__rm/explorer/list
4nd3r: @nico said " basically supporting bsds - realistically hpux, sunos, aix are not so much around anymore". so resolved. also replaced `-name` with `-regex`.

4nd3r:
> You can also skip executing find both in the explorer and in code.

i tried to come out with more readable solution, but didn't manage :disappointed: 

code duplication isn't evil if it allows code to be readable versus ugly oneliner hacks which your friend have to work through 1 year later. and of course someone else can step in later and submit PR which avoids duplication :grinning: 

imo this type is optimized enough for now. so lets merge? :crossed_fingers:

darko-poljak:
@4nd3r Ack. Agree. I will go through it one more time and if I don't notice something fishy it will be merged.

darko-poljak:
@4nd3r Thanks for contribution.

PR #747, opened 2019-04-02
new type: __check_messages

4nd3r:
continuation from chat...

@darko-poljak until we have something better, lets have this type.

i find this dead simple type very useful. maybe it saves someone's day, too :feelsgood:

### DIFF
--- a/cdist/conf/type/__check_messages/gencode-remote
+++ b/cdist/conf/type/__check_messages/gencode-remote
@@ -1,0 +1,27 @@
+#!/bin/sh -e
+#
+# 2019 Ander Punnar (ander-at-kvlt-dot-ee)
+#
+# This file is part of cdist.
+#
+# cdist is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# cdist is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


cdist/conf/type/__check_messages/man.rst
```diff
@@ -0,0 +1,52 @@
...
+For more information about messages see `cdist messaging <cdist-messaging.html>`_.
+
+For more information about dependencies and execution order see
+`cdist manifest <cdist-manifest.html#dependencies>`_ documentation.
+
+
+REQUIRED PARAMETERS
+-------------------
+pattern
+   Search pattern (passed to ``grep -E``).
```
darko-poljak: @4nd3r I would explicitly write that pattern here is extended regular expression :)

cdist/conf/type/__check_messages/gencode-remote
```diff
@@ -0,0 +1,27 @@
...
+# You should have received a copy of the GNU General Public License
+# along with cdist. If not, see <http://www.gnu.org/licenses/>.
+#
+
+if grep -Eq \
+    "$( cat "$__object/parameter/pattern" )" \
+    "$__messages_in"
+then
+    cat "$__object/parameter/execute" \
+        | tee "$__messages_out"
```
darko-poljak: @4nd3r Should `tee` be optional? :thinking:

4nd3r: no.

```
TEE(1)

NAME
       tee - read from standard input and write to standard output and files
```
of course, maybe we don't want output in messages...

PR #748, opened 2019-04-03
UFW (Uncomplicated FireWall)

markasoftware:
I have added the following types:

* `__ufw`: Install and optionally enable UFW
* `__ufw_rule`: Add a normal UFW rule, with all of the UFW command line options translated into cdist-friendly parameters.
* `__ufw_default`: Set default policies (very simple).

This is my first time contributing to cdist, so I have a question: Since UFW's command line is idempotent, my types do not bother using explorers to check whether their work needs to be done -- `gencode-remote` always outputs something. An explorer would have to call `ufw status verbose`, which is not substantially faster than just doing the work I want to do. Is this OK?

Thanks for reviewing this!

### DIFF
--- a/cdist/conf/type/__ufw/man.rst
+++ b/cdist/conf/type/__ufw/man.rst
@@ -1,0 +1,59 @@
+cdist-type__ufw(7)
+==================
+
+NAME
+----
+cdist-type__ufw - Install the Uncomplicated FireWall
+
+
+DESCRIPTION
+-----------
+Installs the Uncomplicated FireWall. Most modern distributions carry UFW in their main repositories, but on CentOS this type will automatically enable the EPEL repository.
+
+Some global configuration can also be set with this type.
+
+OPTIONAL PARAMETERS
+-------------------
+state
+    Either "enabled", "running", "present", or "absent". Defaults to "enabled", which registers UFW to start on boot.
+
+logging
+    Either "off", "low", "medium", "high", or "full". Will be passed to `ufw logging`. If not specified, logging level is not modified.
+
+default_incoming
+    Either "allow", "deny", or "reject". The default policy for dealing with ingress packets.
+    
+default_outgoing
+    Either "allow", "deny", or "reject". The default policy for dealing with egress packets.
+    
+default_routed
+    Either "allow", "deny", or "reject". The default policy for dealing with routed packets (passing through this machine).
+    
+
+EXAMPLES
+--------
+
+.. code-block:: sh
+
+    # Install UFW
+    __ufw
+    # Setup UFW with maximum logging and no restrictions on routed packets.
+    __ufw --logging full --default_routed allow
+
+
+SEE ALSO
+--------
+:strong:`ufw`\ (8)
+
+
+AUTHORS
+-------
+Mark Polyakov <mark@markasoftware.com>
+
+
+COPYING
+-------
+Copyright \(C) 2019 Mark Polyakov. You can redistribute it
+and/or modify it under the terms of the GNU General Public License as
+published by the Free Software Foundation, either version 3 of the
+License, or (at your option) any later version.
--- a/cdist/conf/type/__ufw_rule/gencode-remote
+++ b/cdist/conf/type/__ufw_rule/gencode-remote
@@ -1,0 +1,45 @@
+#!/bin/sh -e
+#
+# 2019 Mark Polyakov (mark@markasoftware.com)
+#
+# This file is part of cdist.
+#
+# cdist is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# cdist is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with cdist. If not, see <http://www.gnu.org/licenses/>.
+#
+
+# This type does not bother with checking the current state of the rules.
+# While it is possible to retrieve the list of rules in a consistent format from
+# `ufw status`, it is a completely different format than the one used on the
+# command line. I also do not suspect it is any faster.
+
+ufw='ufw --force rule'
+
+case "$(cat "$__object/parameter/state")" in
+    present) ;;
+    absent)
+        ufw="$ufw delete"
+        ;;
+    *)
+        echo 'State must be "present" or "absent".' >&2
+        exit 1
+        ;;
+esac
+
+if [ -f "$__object/parameter/rule" ]; then
+    ufw="$ufw $(cat "$__object/parameter/rule")"
+else
+    ufw="$ufw allow $__object_id"
+fi
+
+echo "$ufw"
--- a/cdist/conf/type/__ufw/gencode-remote
+++ b/cdist/conf/type/__ufw/gencode-remote
@@ -1,0 +1,62 @@
+#!/bin/sh -e
+#
+# 2019 Mark Polyakov (mark--@--markasoftware.com)
+#
+# This file is part of cdist.
+#
+# cdist is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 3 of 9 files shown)


cdist/conf/type/__ufw/man.rst
```diff
@@ -0,0 +1,46 @@
...
+
+AUTHORS
+-------
+Mark Polyakov <mark@markasoftware.com>
+
+
+COPYING
+-------
+Copyright \(C) 2019 Mark Polyakov. Free use of this software is
+granted under the terms of the GNU General Public License version 3 (GPLv3).
```
darko-poljak: Can you please use `COPYING` the same as other official types, if you agree? See e.g. `__file` type
(`either version 3 of the License, or (at your option) any later version.`).
Thanks!

darko-poljak:
@4nd3r You might be interested in this type. Wanna review? :)

markasoftware:
After opening this I have realized there are a few changes I want to make: removing __ufw_default and including its functionality as --default_incoming, etc on __ufw, and adding a __ufw_route type which would be mostly copy paste from ufw rule, but for routes that do not terminate on the local machine (ufw has a slightly different syntax for these).

darko-poljak:
@markasoftware Ok, go ahead and improve your type :)

4nd3r:
@darko-poljak i'm from `iptables` generation and i have no experience with ufw :(

but i will skim through a bit...

4nd3r:
would it be possible to generate all rules before adding them with ufw and then diff them with currently active rules? i, myself, wouldn't wrap ufw syntax into cdist parameters, but just use multiple `--rule` parameters which contains ufw syntax.

so, for example, you want to add rules foo, bar, baz and current active rules are foo and bar - rule baz will be added. or you want to add rules foo, bar and active rules are foo, bar, baz - baz will be removed.

```
active rules before: foo, bar
run: __ufw --rule 'foo'  --rule 'bar' --rule 'baz'
active rules after: foo, bar, baz
```
or
```
active rules before: foo, bar, baz
run: __ufw --rule 'foo' --rule 'bar'
active rules after: foo, bar
```

but this way you can only define ufw once per configuration run, because if you define ufw twice, then last ufw will remove first one...

darko-poljak:
> would it be possible to generate all rules before adding them with ufw and then diff them with currently active rules? i, myself, wouldn't wrap ufw syntax into cdist parameters, but just use multiple `--rule` parameters which contains ufw syntax.
> 
> so, for example, you want to add rules foo, bar, baz and current active rules are foo and bar - rule baz will be added. or you want to add rules foo, bar and active rules are foo, bar, baz - baz will be removed.
> 
> ```
> active rules before: foo, bar
> run: __ufw --rule 'foo'  --rule 'bar' --rule 'baz'
> active rules after: foo, bar, baz
> ```
> 
> or
> 
> ```
> active rules before: foo, bar, baz
> run: __ufw --rule 'foo' --rule 'bar'
> active rules after: foo, bar
> ```
> 
> but this way you can only define ufw once per configuration run, because if you define ufw twice, then last ufw will remove first one's rules.

There could be an option to leave old rules or remove them, where by default they won't be removed.
I also agree with `--rule` multiple parameter, instead of wrapping all ufw options in cdist type.

markasoftware:
As soon as we put it in `__ufw` it becomes impossible for rules to be added in multiple locations.

> There could be an option to leave old rules or remove them, where by default they won't be removed.
> I also agree with --rule multiple parameter, instead of wrapping all ufw options in cdist type.

I may be wrong, but I was under the impression that if you attempt to instantiate the same object multiple times, all invocations after the first will be ignored regardless of how the type is written. We also have `__firewalld_rule` and `__iptables_rule`, not `__firewalld` and `__iptables`. Each rule is conceptually a separate "object" IMO.

I agree that wrapping UFW's syntax may be unnecessary though. If we follow `__iptables_rule`'s lead, I could modify `__ufw_rule` so that the only parameter is `rule`, in UFW's format.

> would it be possible to generate all rules before adding them with ufw and then diff them with currently active rules?

Current rules can be listed with `ufw status` or `ufw show added`. Neither necessarily shows rules in the exact same format as they were entered on the command line -- `ufw status` has a completely different format, and `ufw show added` removes unnecessary parts of the command line. Either way, converting between these formats and the CLI format to create a new rule is not trivial.

I think the proper solution to the diffing problem would be "singleton explorers" which run only once for a type, not once per object. Then a type like `__ufw_rule` would have a singleton explorer like `rules_is` which gathers the list of rules at the very start, before any of the gencodes run. Then each individual `__ufw_rule` gencode could check against that explorer very quickly, without running it again. Unfortunately, cdist does not support this (yet!). It might be possible to hack it together with messaging, though.

To summarize, the options are:

* Use `__ufw` and add an optional multiple parameter `--rule`. Pros: better performance. Cons: Hard to convert between rule and status syntax, non-idiomatic (rules are logically separate objects). I don't like it.
* Keep things as is. Pros: It works. Cons: Unnecessary wrapping of UFW syntax, relatively slow when rules already exist because it tries to add them unconditionally.
* Keep separate `__ufw_rule` type, but take a `--rule` parameter that just accepts the command line syntax for the UFW rule instead of the current hodgepodge of parameters. Pros: Follows the precedent set by `__firewalld_rule` and `__iptables_rule` and simple for users who are used to UFW's cli. Cons: Relatively slow when rules already exist.

I am fine with removing the custom parameters, but I strongly believe `__ufw_rule` should be separate from `__ufw`.

cdist/conf/type/__ufw/man.rst
markasoftware: Oops, used @asteven 's script to create the type boilerplate and it had this in it!

darko-poljak:
@markasoftware Since I don't know ufw, I don't have arguments for the way this type is implemented, or how it should be implemented. I will trust you.
@markasoftware Do you have some changes in mind?

@telmich @asteven Do you have any remarks/suggestions?

markasoftware:
After reading your comments and other types, I plan to change ufw rule to not wrap ufw syntax.

markasoftware:
I have removed the syntax wrapping in `__ufw_rule`. Unless there are further comments I think this is ready to merge.

cdist/conf/type/__ufw/man.rst
```diff
@@ -0,0 +1,59 @@
+cdist-type__ufw(7)
+================
```
darko-poljak: @markasoftware To avoid sphinx warnings (iirc) `=` line length should be equal to text length.

cdist/conf/type/__ufw_rule/gencode-remote
```diff
@@ -0,0 +1,45 @@
...
+        ufw="$ufw delete"
+        ;;
+    *)
+        echo 'State must be "present" or "absent".' >&2
+        exit 1
+        ;;
+esac
+
+if [ -f "$__object/parameter/rule" ]; then
+    ufw="$ufw $(cat "$__object/parameter/rule")"
```
darko-poljak: @markasoftware Can `--rule` parameter be multiple? So you can use it multiple times at command line for multiple rules.

darko-poljak: @markasoftware Hm... `__firewalld_rule` and `__iptables_rule` types don't have `--rule` parameter defined as multiple so, to be consistent, this is fine.
Sorry for the inconvenience :)

cdist/conf/type/__ufw/man.rst
darko-poljak: Only this little thing.

And can you rebase your branch to have only one commit?
For nicer history :)

markasoftware: Sure. I would have liked to split into two commits (one for __ufw and another for __ufw_rule) but didn't bother.

markasoftware: Also, you as a maintainer can do this via the GitHub UI when merging a PR: https://help.github.com/en/articles/about-pull-request-merges#squash-and-merge-your-pull-request-commits

darko-poljak: @markasoftware Please resolve sphinx warning here:
```
cdist/docs/src/man7/cdist-type__ufw.rst:2: WARNING: Title underline too short.

cdist-type__ufw(7)
================
```

markasoftware: Done!

PR #749, opened 2019-04-05
Add debug dump script

darko-poljak:
Hi!

@4nd3r has expressed the need for something that dumps outputs, and that can be used for debugging.
This script dumps data from cache directory, for specified hosts or for all it finds in cache directory.

It can dump the following outputs:
- all
- code-*
- global explorers
- type explorers
- gencode-*
- messages
- captured stdout
- captured stderr
- parameters.

By default, all lines are prefixed with filename (this can be turned off).
One can also turn on line number prefix.

@4nd3r @telmich @asteven Do you find this script useful?
Do you have suggestions for its improvement?

First, I was thinking about whether or not to implement this support in cdist core.
But since its purpose is only for debugging, and since cdist already dumps data to cache directory,
I came up with this helper script.

### DIFF
--- a/docs/src/cdist-troubleshooting.rst
+++ b/docs/src/cdist-troubleshooting.rst
@@ -43,3 +43,21 @@ you write to use the -e flag:
     % cat ~/.cdist/manifest/special
     #!/bin/sh -e
     ...
+
+Using debug dump helper script
+------------------------------
+Since cdist stores data to local cache that can be used for debugging there
+is a helper script that dumps data from local cache.
+
+For more info see:
+
+.. code-block:: sh
+
+    debug-dump.sh -h
+
+Or from cdist git cloned directory:
+
+.. code-block:: sh
+
+    ./scripts/debug-dump.sh -h
+
--- a/scripts/debug-dump.sh
+++ b/scripts/debug-dump.sh
@@ -1,0 +1,325 @@
+#!/bin/sh
+
+VERSION="0.0.1"
+RELEASE=""
+
+set -u
+# set -x
+
+hosts=
+cache_dir=~/.cdist/cache
+
+do_all=1
+do_global_explorer=
+do_type_explorer=
+do_script_stdout=
+do_script_stderr=
+do_gencode=
+do_code=
+do_messages=
+do_parameter=
+delimiter=':'
+ln=
+filename_prefix=1
+verbose=0
+
+myname=${0##*/}
+
+print_version()
+{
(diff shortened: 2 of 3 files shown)


asteven:
Awesome! This will be useful!
I agree that helper script is the way to go.
Maybe some white space between file name prefix and content would be nice? Not sure.

darko-poljak:
@asteven I rely on `grep -H` for filename prefix. Also, on `-n` for line number prefix.
`find ... -exec grep [-n] [-H] '' {} ...`.

darko-poljak:
Although I now see that `-H` is not by POSIX.

darko-poljak:
@asteven I switched from grep to awk, so now we have control on the output :)

darko-poljak:
@asteven And something like the following? (`|` after `:`, if `:` is even needed).

```target host: cdisttesthost, host directory: ./759547ff4356de6e3d9e08522b0d0807
==============================================================================
./explorer/disks:|1:|sd? hd? vd?
./explorer/hostname:|1:|icosahedron
./explorer/init:|1:|init
./explorer/interfaces:|1:|em0
./explorer/interfaces:|2:|lo0
./explorer/interfaces:|3:|wlan0
./explorer/kernel_name:|1:|FreeBSD
./explorer/machine:|1:|amd64
./explorer/machine_type:|1:|unknown
./explorer/os:|1:|freebsd
./explorer/os_version:|1:|11.2-RELEASE-p9
./object/__dryrun_test/testit/.cdist-gm7q2boc/code-local:|1:|echo "This gencode-local script should never be executed >&2"
./object/__dryrun_test/testit/.cdist-gm7q2boc/code-local:|2:|exit 1
./object/__dryrun_test/testit/.cdist-gm7q2boc/code-remote:|1:|echo "this gencode-remote script should never be executed >&2"
./object/__dryrun_test/testit/.cdist-gm7q2boc/code-remote:|2:|exit 1
./stderr/remote:|1:|cat: /etc/os-release: No such file or directory
./stderr/remote:|2:|cat: /usr/lib/os-release: No such file or directory
```

4nd3r:
> (| after :, if : is even needed)

just regular space after `<file>:` and it's fine imho.

darko-poljak:
@4nd3r @asteven I think it is the best to support delimiter option, with current `:` as the default :)

darko-poljak:
@asteven @telmich @4nd3r I also mentioned this helper script in cdist docs.

If you don't have any other remarks or suggestions then I will merge this.

4nd3r:
we can always change things later :thought_balloon:

PR #753, opened 2019-04-11
fix disks explorer

4nd3r:
```
$ cdist-dump -E test | grep disks
./explorer/disks:sda sdb sdc sdd sde hd? vd?
```
bug, because `hd?` and `vd?` are not expanded.

### DIFF
--- a/cdist/conf/explorer/disks
+++ b/cdist/conf/explorer/disks
@@ -11,7 +11,7 @@ case "$os" in
     ;;
     *)
         # hopefully everything else is linux
-        if which lsblk > /dev/null
+        if command -v lsblk > /dev/null
         then
             # exclude ram disks and cdroms 
             lsblk -e 1,11 -dno name | xargs
--- a/cdist/conf/explorer/disks
+++ b/cdist/conf/explorer/disks
@@ -13,7 +13,7 @@ case "$os" in
         # hopefully everything else is linux
         if command -v lsblk > /dev/null
         then
-            # exclude ram disks and cdroms 
+            # exclude ram disks, floppies and cdroms 
             lsblk -e 1,2,11 -dno name | xargs
         else
             # fallback
--- a/cdist/conf/explorer/disks
+++ b/cdist/conf/explorer/disks
@@ -14,7 +14,7 @@ case "$os" in
         if command -v lsblk > /dev/null
         then
             # exclude ram disks and cdroms 
-            lsblk -e 1,11 -dno name | xargs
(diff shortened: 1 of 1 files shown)


darko-poljak:
@telmich @asteven You two should review this.

cdist/conf/explorer/disks
```diff
@@ -1,16 +1,26 @@
...
     ;;
-
     *)
-        cd /dev || exit 0
-        echo sd? hd? vd?
+        # hopefully everything else is linux
+        if command -v lsblk > /dev/null
+        then
+            # exclude ram disks and cdroms 
+            lsblk -e 1,11 -dno name | xargs
```
asteven: Please also exclude floppy disks.

```suggestion
            lsblk -e 1,2,11 -dno name | xargs
```

cdist/conf/explorer/disks
```diff
@@ -1,16 +1,26 @@
...
+        else
+            # fallback
+            find /dev \
+                -type b \
+                -regex '/dev/\(hd\|sd\|vd\)[a-z]+' \
```
asteven:
```suggestion
                -regex '/dev/\(fd\|hd\|sd\|vd\)[a-z]+' \
```

4nd3r: `-regex` here is inclusive. which one you want? :smile_cat:

darko-poljak:
@4nd3r Do you also have NetBSD? It seems that NetBSD also supports `sysctl hw.disknames`.
http://netbsd.gw.com/cgi-bin/man-cgi?sysctl+7+NetBSD-current

It would be nice to add support for it too, since there is support for OpenBSD and FreeBSD :)

4nd3r:
yes, i have. will check.

4nd3r:
at least that netbsd i have running uses `ld*` naming. not sure if there are other possibilities...

darko-poljak:
> at least that netbsd i have running uses `ld*` naming. not sure if there are other possibilities...

@4nd3r  Hm.... It seems ld is just for SSD.
https://wiki.netbsd.org/Installation_on_UEFI_systems/
See `Identifying Disks`.

4nd3r:
thanks for the link.

> Depending on details a SSD will show up as ld (NVME device) or wd (SATA device). Hard disks usually show up as wd as well. Other disks may show up as ld or sd (SCSI, SAN, RAID, ...).

ld, wd and sd should be enough. dk-s are partitions.

darko-poljak:
@telmich @asteven Any new remarks/suggestions?

PR #755, opened 2019-04-12
fix __(file|directory)/explorer/stat for BSDs

4nd3r:
for example: %Op = 100644 vs %OLp = 644

### DIFF
--- a/cdist/conf/type/__directory/explorer/stat
+++ b/cdist/conf/type/__directory/explorer/stat
@@ -25,21 +25,13 @@ destination="/$__object_id"
 
 os=$("$__explorer/os")
 case "$os" in
-   "freebsd"|"netbsd"|"openbsd")
-      # FIXME: should be something like this based on man page, but can not test
-      stat -f "type: %ST
+   "freebsd"|"netbsd"|"openbsd"|"macosx")
+      stat -f "type: %HT
 owner: %Du %Su
 group: %Dg %Sg
-mode: %Op %Sp
-" "$destination"
+mode: %Lp %Sp
+" "$destination" | awk '/^type/ { print tolower($0); next; } { print; }'
    ;;
-    "macosx")
-       stat -f "type: %HT
- owner: %Du %Su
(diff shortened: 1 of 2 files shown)


4nd3r:
do not merge yet.

darko-poljak:
@4nd3r Perhaps you can add some title prefix, like `Not For Merge Yet:` (gitlab has `WIP` (work in progress)). And when you remove that prefix we know the code is ready for final review and merge.

4nd3r:
not sure if problem.

```
root@ander-test-openbsd:~# stat -f '%HT' .bashrc
Regular File
root@ander-test-debian:~# stat --printf='%F\n' .bashrc
regular file
```

bsd stat return file type as `Upper Case First Letter`, linux all lower case.

darko-poljak:
I think that explorer should return the same values, regardless of the OS.
So the easier would be to translate all to lower case.

darko-poljak:
tr '[:upper:]' '[:lower:]'

4nd3r:
we can't do it with entire output, because some strange people might use CamelCase users/groups :smile:

4nd3r:
damn... bsd awk sucks :disappointed:

darko-poljak:
Why?

4nd3r:
i'm too used with gnu awk. same with sed :sob:

darko-poljak:
That is why I usually only use what POSIX defines ;)

4nd3r:
coworker helped me out here. @darko-poljak are you fine with this "hack" ?

darko-poljak:
> coworker helped me out here. @darko-poljak are you fine with this "hack" ?

No.

I think ` | awk '/^type/ { print tolower($0); next; } { print; }'` is more elegant.

4nd3r:
thanks @darko-poljak 

and imho this PR is ready now. should I rebase?

darko-poljak:
@4nd3r Yes, please.

4nd3r:
and done

PR #759, opened 2019-04-15
Add support for ACL mask.

jimis:


cdist/conf/type/__acl/explorer/acl_is
```diff
@@ -19,5 +19,8 @@
 #
 
 if [ -e "/$__object_id" ]
-then getfacl "/$__object_id" | grep -E '^((default:|)(user|group)):[a-z]' || true
+then
+    getfacl "/$__object_id" \
+        | grep -E '(^((default:|)(user|group)):[a-z])|(^(default:|)mask::[a-z])' \
```
4nd3r: after few months i see that i have been lazy.

i would go with `^((default:)?(user|group):[^:]|(default:)?mask::)`

cdist/conf/type/__acl/gencode-remote
```diff
@@ -37,7 +37,19 @@ do
...
+    while read -r l
+    do
+        echo "mask::$l"
+
+        if [ -f "$__object/parameter/default" ]
+        then echo "default:mask::$l"
+        fi
+    done < "$__object/parameter/mask"
+fi
+)"
```
4nd3r: there can be only one mask per file, so no while loop needed here?

cdist/conf/type/__acl/parameter/optional_multiple
```diff
@@ -1,2 +1,3 @@
 user
 group
+mask
```
4nd3r: this should be in `parameter/optional` then

4nd3r:
I also noticed two mistakes made by me - strange that @darko-poljak didn't notice them.

first: https://github.com/ungleich/cdist/blob/806b84444b04d6def9b168437a43517c041284a1/cdist/conf/type/__acl/gencode-remote#L58
second: https://github.com/ungleich/cdist/blob/806b84444b04d6def9b168437a43517c041284a1/cdist/conf/type/__acl/gencode-remote#L68

first should be `-Eq` and second `-Fq', because we only care about exit code.

@jimis, can you fix those line too, when you are already at it? :smile_cat:

jimis:
OK I'll test first and re-submit the pull request afterwards.

jimis:
Rebased my first commit to include your suggestions and added the `grep -q` fix on a 2nd commit.

cdist/conf/type/__acl/man.rst
```diff
@@ -21,6 +21,9 @@ user
 group
    Add group ACL entry.
 
+mask
+   Add mask ACL entry.
```
4nd3r: this should be in "Optional" section :smiling_imp:

4nd3r:
actually, must check one thing...

4nd3r:
https://github.com/ungleich/cdist/blob/933eef3e21b37f57cd006c500c905a158caa0870/cdist/conf/type/__acl/explorer/acl_is#L23

@jimis I once more abuse of your kindness :smiley_cat:  this should be `setfacl -p`, otherwise stderr contains "getfacl: Removing leading '/' from absolute path names". this only changes first comment line of output, but we don't use it.

```
root@ander-test-debian:~# getfacl -p /tmp/cdist-test > test1
root@ander-test-debian:~# getfacl /tmp/cdist-test > test2
getfacl: Removing leading '/' from absolute path names
root@ander-test-debian:~# diff -u test1 test2
--- test1       2019-04-15 18:58:48.085901501 +0300
+++ test2       2019-04-15 18:58:52.785914706 +0300
@@ -1,4 +1,4 @@
-# file: /tmp/cdist-test
+# file: tmp/cdist-test
 # owner: root
 # group: root
 user::rwx
```

cdist/conf/type/__acl/man.rst
jimis: Fixed, amended the 1st commit.

jimis:
@4nd3r AFAICT `getfacl -p` is very linux-specific. I don't see it in the man pages of OpenBSD, FreeBSD, SunOS, IllumOS, SmartOS.

4nd3r:
hmm, then we should use `getfacl ... 2>/dev/null` to suppress the output to stderr.

@darko-poljak what do you think?

4nd3r:
and openbsd actually don't have `(get|set)facl` :smile: 

@darko-poljak lets merge this PR and I'll improve `__acl` in another PR.

darko-poljak:
@4nd3r @jimis yes, `getfacl ... 2>/dev/null`

jimis:
Fixed. 

@4nd3r indeed OpenBSD doesn't have `getfacl`, my mistake, I was looking in FreeBSD's man page but from OpenBSD's website.

telmich:
Wow. you are all **crazy**. Not that ACLs by design are highly complex, as is the implementation and you even managed to add a type for it.

I am **NOT** going to review this one (keeping my sanity) and leaving full responsibility to @darko-poljak .

ACLs... wow... this is Windows NT4 talk coming back... in any case, wish you all together good luck with your sanity ;-)

4nd3r:
@telmich you n00b, `acl` conforms to POSIX 1003.1E draft standard 17, since '99! :smiling_imp:

darko-poljak:
> Wow. you are all **crazy**. Not that ACLs by design are highly complex, as is the implementation and you even managed to add a type for it.
> 
> I am **NOT** going to review this one (keeping my sanity) and leaving full responsibility to @darko-poljak .
> 
> ACLs... wow... this is Windows NT4 talk coming back... in any case, wish you all together good luck with your sanity ;-)

And I will rely on @4nd3r and @jimis :)

darko-poljak:
@4nd3r I am now merging this one, so you can prepare new PR.

jimis:
Thanks for merging @4nd3r @darko-poljak !

@telmich I had one problem, and decided to use ACLs to solve it. And now I have two problems... :-)

darko-poljak:
Ha ha ha ha ha

Darko

PR #763, opened 2019-04-19
explorer/disks: do fallback right, in a POSIX way

darko-poljak:
https://github.com/ungleich/cdist/commit/aba1ae68f073382d7f0488b98c69af3e46b8d26a is wrong!

I don't have hd, sd, nor vd and the result is:
```
$ cd /dev && echo [vsh]d?
[vsh]d?
```
And https://github.com/ungleich/cdist/commit/1c152f0acbf2f7531fd4c129b04a1658654aa05e does not work with `find` implementations without `-printf`.

So here I am trying to do it with tools that POSIX specifies.

@telmich Can you test this one with alpine linux?
@4nd3r Can you test this with your machines if you have some for fallback?

### DIFF
--- a/cdist/conf/explorer/disks
+++ b/cdist/conf/explorer/disks
@@ -1,27 +1,27 @@
 #!/bin/sh -e
 
-os="$( "$__explorer/os" )"
+uname_s="$(uname -s)"
 
-case "$os" in
-    freebsd)
+case "${uname_s}" in
+    FreeBSD)
         sysctl -n kern.disks
     ;;
-    openbsd)
-        sysctl -n hw.disknames | grep -Eo '[sw]d[0-9]+' | xargs
+    OpenBSD|NetBSD)
+        sysctl -n hw.disknames | grep -Eo '[lsw]d[0-9]+' | xargs
     ;;
-    netbsd)
-        sysctl -n hw.disknames | grep -Eo '[lsw]d[0-9]' | xargs
-    ;;
-    *)
-        # hopefully everything else is linux
+    Linux)
         if command -v lsblk > /dev/null
         then
             # exclude ram disks, floppies and cdroms
             # https://www.kernel.org/doc/Documentation/admin-guide/devices.txt
             lsblk -e 1,2,11 -dno name | xargs
         else
-            # fallback
-            cd /dev && echo [vsh]d?
+            printf "%s operating system without lsblk is not supported, if you can please submit a patch\n" "${uname_s}" >&2
+            exit 1
         fi
+    ;;
+    *)
+        printf "%s operating system is not supported, if you can please submit a patch\n" "${uname_s}" >&2
+        exit 1
     ;;
 esac


telmich:
Why so complicated? What about `cd /dev; find . | awk '/(v|h|s)d?/ { print $0 }'
(written on the way in the mountains, take a bit of salt with it)

darko-poljak:
> Why so complicated? What about `cd /dev; find . | awk '/(v|h|s)d?/ { print $0 }'
> (written on the way in the mountains, take a bit of salt with it)

@telmich This prints all in `/dev` that contains two characters where first of two is 'v' or 'h' or 's'.

Is sufix for `(v|h|s)d` always one letter?

darko-poljak:
We also need it in one line.

And no `./` prefix in names.

And no names in subdirectories.

darko-poljak:
I think we can shorten it into:

```
cd /dev && find . | awk 'BEGIN{ORS=" "} /^\.\/(h|s|v)d[a-z]+$/ {sub("./", ""); print; }'
```

4nd3r:
> Is sufix for (v|h|s)d always one letter?

no. 

https://www.kernel.org/doc/Documentation/admin-guide/devices.txt

SCSI disks can go beyong one letter suffix.

for example, machine with lots of LUNs behind multipath:

```
root@X:/dev# echo [vsh]d?
sda sdb sdc sdd sde sdf sdg sdh sdi sdj sdk sdl sdm sdn sdo sdp sdq sdr sds sdt sdu sdv sdw sdx sdy sdz
root@X:/dev# find . | awk 'BEGIN{ORS=" "} /^\.\/(h|s|v)d[a-z]+$/ {sub("./", ""); print; }'; echo
sdap sdao sdbc sdbb sdba sdaz sday sdax sdaw sdav sdan sdak sdaj sdaf sdae sdad sdab sdz sdl sdk sdi sdg sdf sde sdd sdb sda sdas sdar sdaq sdam sdal sdai sdah sdag sdac sdaa sdx sdw sdv sdu sdt sds sdr sdq sdp sdo sdn sdm sdj sdh sdc sdy 
```

4nd3r:
```
# uname -a
SunOS ander-test-openindiana 5.11 illumos-42a3762d01 i86pc i386 i86pc
# cd /dev && find . | awk 'BEGIN{ORS=" "} /^\.\/(h|s|v)d[a-z]+$/ {sub("./", ""); print; }'
awk: syntax error near line 1
awk: illegal statement near line 1
```

darko-poljak:
Can you resolve this? What is illegal?

Darko

darko-poljak:
> ```
> # uname -a
> SunOS ander-test-openindiana 5.11 illumos-42a3762d01 i86pc i386 i86pc
> # cd /dev && find . | awk 'BEGIN{ORS=" "} /^\.\/(h|s|v)d[a-z]+$/ {sub("./", ""); print; }'
> awk: syntax error near line 1
> awk: illegal statement near line 1
> ```

Which part of awk script is wrong?

telmich:
probably ok :-)

4nd3r:
as @asteven said - maybe supporting ancient operating systems isn't worth an effort.

and instead of `os` explorer, shouldn't we use `kernel_name` or something which allows us to use one universal label for linux distros? fallback with assumption that everything else is linux isn't actually good idea, imho.

darko-poljak:
I found a problem, we should refactor this explorer. https://github.com/ungleich/cdist/commit/dbcf7d55432c90bb6be4547d6e49a1e060279a9a introduced dependency on os explorer, since it uses its value. Well, os and disks are global explorers and there is no support for order of execution for global explorers.

darko-poljak:
There are other global explorer that rely on os explorer value. How does this work?

4nd3r:
it works if you use absolute path

```
ander@limukas:~/Desktop$ cat test1.sh 
#!/bin/sh -e
echo test
ander@limukas:~/Desktop$ cat test2.sh 
#!/bin/sh -e
x="$( './test1.sh' )"
echo "$x"
ander@limukas:~/Desktop$ ./test2.sh 
test
```

so everything is correct - global explorers execute global explorers they need

4nd3r:
i would go with something like this:

```
case "$( uname -s )" in
    FreeBSD)
        sysctl -n kern.disks
    ;;
    OpenBSD)
        sysctl -n hw.disknames | grep -Eo '[sw]d[0-9]+' | xargs
    ;;
    NetBSD)
        sysctl -n hw.disknames | grep -Eo '[lsw]d[0-9]' | xargs
    ;;
    Linux)
        if command -v lsblk > /dev/null
        then
            # exclude ram disks, floppies and cdroms
            # https://www.kernel.org/doc/Documentation/admin-guide/devices.txt
            lsblk -e 1,2,11 -dno name | xargs
        else
            # fallback
            echo ''
        fi
    ;;
    *)
        echo 'your OS is not supported' >&2
    ;;
esac
```

telmich:
you can use $__global/explorer to execute explorers. See os_version.


Ander Punnar <notifications@github.com> writes:

> i would go with something like this:
>
> ```
> case "$( uname -s )" in
>     FreeBSD)
>         sysctl -n kern.disks
>     ;;
>     OpenBSD)
>         sysctl -n hw.disknames | grep -Eo '[sw]d[0-9]+' | xargs
>     ;;
>     NetBSD)
>         sysctl -n hw.disknames | grep -Eo '[lsw]d[0-9]' | xargs
>     ;;
>     Linux)
>         if command -v lsblk > /dev/null
>         then
>             # exclude ram disks, floppies and cdroms
>             # https://www.kernel.org/doc/Documentation/admin-guide/devices.txt
>             lsblk -e 1,2,11 -dno name | xargs
>         else
>             # fallback
>             echo ''
>         fi
>     ;;
>     *)
>         echo 'your OS is not supported' >&2
>     ;;
> esac
> ```


--
Your Swiss, Open Source and IPv6 Virtual Machine. Now on www.datacenterlight.ch.

darko-poljak:
Ooohhhh! I think I need rest, a vacation. I didn't noticed command substitution and forgotten what __explorer env var is. Sorry for the inconvenience.

4nd3r:
[go here](https://loodusegakoos.ee/where-to-go/recreation-areas/rapina-varska-recreation-area/koverajarve-rental-hut)

image

darko-poljak:
@telmich @4nd3r My final proposal. If you agree then we can merge it and finally release new version.
Beside this, no more patches before new release.

4nd3r:
imho go for it

PR #767, opened 2019-04-20
Report explorer error better

### LINKED ISSUE #766: stderr from explorers
in global and type explorers, when `exit 1` with `echo error >&2` is emitted, stderr of explorer is not shown, but it is with gencode-remote:

```
$ cdist config -i foo bar
INFO: [3112]: bar: Starting configuration run
ERROR: [3112]: bar: /home/ander/Desktop/cdist/cdist/conf/type/__acl/gencode-remote: ['/home/ander/Desktop/cdist/cdist/conf/type/__acl/gencode-remote']

Error processing object '__acl/tmp/cdist-test'
==============================================
name: __acl/tmp/cdist-test
path: /tmp/tmpt_ifzwx0/bca1c5264040cbe26816461c74f822d3/data/object/__acl/tmp/cdist-test/.cdist-q9f4wtul
source: acl-add
type: /home/ander/Desktop/cdist/cdist/conf/type/__acl

gencode-remote:stderr
---------------------
getfacl not available
```

also, since type explorer fails, nothing is saved into cache directory, so I can't even use cdist-dump.

usecase: I would like to do preflight checks in type explorer and if something is wrong, type should terminate with exit 1 and error show to admin. workaround right now is collect messages into output of type explorer and do the check in gencode-remote.

darko-poljak:
Fixes #766

@4nd3r Can you test it and see if anything is missing or can be improved?

@asteven @telmich Take a look.

### DIFF
--- a/cdist/__init__.py
+++ b/cdist/__init__.py
@@ -181,22 +181,59 @@ class CdistObjectError(CdistEntityError):
                          params, stdout_paths, stderr_paths, subject)
 
 
+class CdistObjectExplorerError(CdistEntityError):
+    """
+    Something went wrong while working on a specific
+    cdist object explorer
+    """
+    def __init__(self, cdist_object, explorer_name, explorer_path,
+                 stderr_path, subject=''):
+        params = [
+            ('object name', cdist_object.name, ),
+            ('object path', cdist_object.absolute_path, ),
+            ('object source', " ".join(cdist_object.source), ),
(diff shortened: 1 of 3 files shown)


4nd3r:
thanks @darko-poljak - that's perfect!

darko-poljak:
@4nd3r Does this mean you have already tried it?

4nd3r:
yes, i tried

PR #776, opened 2020-05-27
fixup os_version explorer for sles 15

dheule:
Suse has no more SuSE-release as of SLES15, so the os explorer takes the os-release in to account:
`NAME="SLES"
VERSION="15-SP1"
VERSION_ID="15.1"
PRETTY_NAME="SUSE Linux Enterprise Server 15 SP1"
ID="sles"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:15:sp1"
`
Since there the ID= is checked, the result of os is no more suse, its now sles.

### DIFF
--- a/cdist/conf/explorer/os_version
+++ b/cdist/conf/explorer/os_version
@@ -60,7 +60,7 @@ case "$("$__explorer/os")" in
    slackware)
       cat /etc/slackware-version
    ;;
-   suse)
+   suse|sles)
       if [ -f /etc/os-release ]; then
         cat /etc/os-release
       else


sideeffect42:
@dheule What is is your opinion on https://code.ungleich.ch/ungleich-public/cdist/issues/785?

4nd3r:
:thinking:

since I'm not suse-user, I can't pick sides. if you say that sles != suse, then we need to distinguish between them, but if they are essentially the same, then no. but same time we do distinguish redhat and its derivatives...

@telmich what's your say?

telmich:
If there is no difference for any affected types, keep sles = suse.

sideeffect42:
I'm not an enterprise Linux user, either, which is why I thought that we should ask someone who uses SuSE.  
AFAICT SLES was reported as `suse` until they dropped `/etc/SuSE-release` with SLES 15(?).

So if the behaviour should not change, all SuSE variations should be reported as `suse`.

But @4nd3r brought up a good point about RedHat.
Its variations have historically been differentiated in cdist. Same about Debian/Devuan/(Ubuntu).  
Illumos on the other hand is reported as `solaris` even though there are clear differences between the two.

I'd argue that for almost all types the minor differences between the variations of very similar distros can be ignored, yet I can imagine that there may be use cases where the differences matter.  
What are your thoughts on having two explorers? A more coarse one and a finer one.

4nd3r:
ubuntu, debian and devuan have differences, same with redhat, centos and fedora. solaris is another story... I don't scratch that itch anymore, but if you ask me, solaris and illumos must be distinguished. but for that we have to wait PRs from someone who is actually using solaris-ish.

so, the question still is: are sles and suse so different that we need to distinguish them or maybe just make sles report as suse? did some quick googling and looks like sles is subset of suse, basically stable enterpise branch. maybe i got wrong impression :shrug: 

tho, we should trust actual users here.

@dheule what's your guts feeling about other types which have "suse" string in them? someone should catch up with them too.

dheule:
My opinion is that sles and suse is realy near in most aspects. 

So instead of merging my commits, it's also ok to change the os explorer in some way to output suse.

Best regards,
Daniel

dheule:
If we take https://github.com/ungleich/cdist/pull/778, cancel this one ...

4nd3r:
closed in favor of #778

PR #780, opened 2021-10-30
Handle missing apt lists

contradict:
On a newly imaged system, /var/lib/apt/lists may not exist. Check for
this case before comparing dates to the sources.

### DIFF
--- a/cdist/conf/type/__apt_update_index/gencode-remote
+++ b/cdist/conf/type/__apt_update_index/gencode-remote
@@ -20,7 +20,7 @@
 
 # run 'apt-get update' if anything in /etc/apt is newer then /var/lib/apt/lists
 cat << DONE
-if find /etc/apt -mindepth 1 -cnewer /var/lib/apt/lists | grep . > /dev/null; then
+if [ ! -d /var/lib/apt/lists ] || { find /etc/apt -mindepth 1 -cnewer /var/lib/apt/lists | grep . > /dev/null; }; then
    apt-get update || apt-get update
 fi
 DONE


4nd3r:
Thanks.

And while on this, I would also change:

```sh
.. || [ -n "$( find /etc/apt -mindepth 1 -cnewer /var/lib/apt/lists )" ]
```

Same time there has been some discussion in upstream gitlab about different ways of updating APT package index:
https://code.ungleich.ch/ungleich-public/cdist/-/issues/862

tl;dr It's little messy :smile: 

Also gitlab->github syncing has been broken for some time, I'll look into it.

contradict:
How about something like that? Would it be helpful for me to redo this PR on your gitlab instance?

4nd3r:
To be honest, yes. Because project's GitHub presence is kind of broken atm. Tho I understand not wanting to create account just to submit oneliner patch.

And that linked thread in upstream GitLab... there's no consensus yet and it moves slowly. I suggest holding this patch until we have, khm, decision. Feel free to jump in with your ideas and thoughts. Much appreciated. But my vote currently goes to solving everything related to APT index updating in `__package_apt` type. Will see.

Meanwhile, to unblock yourself, you can store your modified type in `~/.cdist/type/` and use it normally in your manifests. 

For example, I have following conf:
```sh
$ grep conf_dir ~/.cdist.cfg 
conf_dir = /home/ander/.cdist:/home/ander/Desktop/src/cdist/cdist/conf:/home/ander/Desktop/src/cdist-contrib
```

This way I can keep my modifications separated from upstream repos and not worry about when/if my PR will be accepted/merged.

PR #781, opened 2021-11-13
__package_apt: Mark already installed packages as manual

sideeffect42:
This PR extends the `__package_apt` type to mark requested packages that were automatically installed e.g. as a dependency for some other package as manually installed.
This prevents these packages from being uninstalled by `apt-get autoremove` should e.g. the reverse dependency be removed or install recommends be disabled.

### DIFF
--- a/cdist/conf/type/__package_apt/explorer/state
+++ b/cdist/conf/type/__package_apt/explorer/state
@@ -1,6 +1,7 @@
 #!/bin/sh
 #
 # 2011-2012 Nico Schottelius (nico-cdist at schottelius.org)
+# 2021 Dennis Camera (cdist at dtnr.ch)
 #
 # This file is part of cdist.
 #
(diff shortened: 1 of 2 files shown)


asteven:
Hi @sideeffect42,

Can you elaborate on what problem this solves for you?

At first sight, it seems like an anti-feature.
I usually want `apt-get autoremove` to do what it does.
Why is this patch needed?

Cheers

asteven:
> Hi @sideeffect42,
> 
> Can you elaborate on what problem this solves for you?
> 
> At first sight, it seems like an anti-feature. I usually want `apt-get autoremove` to do what it does. Why is this patch needed?
> 
> Cheers

Never mind. Read through the code again after coffee and it off course makes sense.

+1

4nd3r:
Thanks, merged.

PR #785, opened 2022-04-19
Extend __unpack cases to support .txz and .tbz2

travispaul:
`cdist/conf/type/__unpack/manifest` already had a case for `.txz` but it was missing from the gencode-local code. Also added .bz2 support while here.

Ran into this as FreeBSD base sets are distributed as `.txz` files.


travispaul:
Going to go ahead and close this as there seems to be no interest and I'm also no longer using cdist.

